PR #14, opened 2025-07-18
Initial work on progress saving

danielw97:
Hi,
Here's my initial draft for document position saving. currently, it saves on tab switch, tab exit, close all tabs, window close and application exit.
the last one is due to the fact that positions weren't saving when the application was exited with alt+f4, so it's essentially to catch that.
If you have any ideas to handle this any better, feel free to suggest or make those changes.
There are a few issues with the current way I've done this, as I see it. if a user switches tabs very quickly, the position will be saved every time.
I'm going to be doing some testing in the nexxt few days of this feature in practice, but wanted to open this at least as a draft so you can have a look and let me know your thoughts.

### DIFF
--- a/src/app.cpp
+++ b/src/app.cpp
@@ -29,7 +29,7 @@ int app::OnExit() {
 				long position = data->textbox->GetInsertionPoint();
 				wxConfigBase* config = wxConfigBase::Get();
 				if (config) {
-					config->SetPath("/DocumentPositions");
+					config->SetPath("/documents");
 					config->Write(data->file_path, position);
 				}
 			}
--- a/src/app.cpp
+++ b/src/app.cpp
@@ -3,6 +3,7 @@
 #include "parser.hpp"
 #include <wx/filename.h>
 #include <wx/stdpaths.h>
+#include <wx/config.h>
 
 bool app::OnInit() {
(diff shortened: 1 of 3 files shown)


trypsynth:
Thanks a ton for this, I'll review it this evening/this weekend!
Out of curiosity, was this vibe-coded? The comments and blank line style seem kinda like Claude. For the record, I'm okay with this. I don't personally vibe code much, but Paperback is by no means a "No-AI project".

danielw97:
No problem at all.
This wasn't vibe coded as such in its entirety, but I did use Claude 4 sonnet as a learning tool when doing this, particularly for the config saving and timer as it was an area I wasn't as familiar with.
I don't code in my day job, so this is very much a learning opportunity for me.
Happy to go over and amend any formatting, if needed. This was a first pass to get this implemented, and we can always refine it further.

trypsynth:
Makes sense, thanks!
For some reason, the GitHub CLI is refusing to let me checkout your branch, gh pr checkout 14 errors with:
> ! [rejected]        refs/pull/14/head -> danielw97/master  (non-fast-forward)           
                               
> failed to run git: exit status 1                                                                                        
From glancing at the diff this looks okay, I'll merge it in and then we can tweak it once in master if needed :)

Issue #11, opened 2025-07-17
add support for saveing the reading position in the book


aryanchoudharypro:
hello, i would like to request the ability to save my reading position in the book when i close the app.

trypsynth:
Hi,
This is certainly planned. I already have a basic ini configuration system, and my thought is to store a file path as the key, the position through the book (in characters) as the value.
There are some things to consider here, for example we should probably only actually write changes to the currently active tab to disk to avoid overhead, but this is certainly planned.
I'm working as a software development intern this summer and that's where I'm allocating most of my programming energy for the next month or so, but I definitely plan to add this and thanks for opening it!

jaybird110127:
I'd also be very interested in this. Also, the ability to add and jump to bookmarks within a document, and have these saved between runs of the program.

trypsynth:
Hi @danielw97 
If you want to tackle it, feel free! Don't worry about stepping on toes in PR's, if something looks not right I'll do a gh pr review, or even just force-push to your branch if git will let me. It is appreciated you checking on how we want to do implementations and definitely keep that up, but a pull request isn't finite and I can always change it, so don't stress that too much.
I think storing them in the config seems reasonable yeah. My thought is to have a [documents] or [books] subsection, with something like C:\Users\Quin\Dropbox\books\authors\Ami Spencer\Broken.epub = 43784.
This does make bookmarks trickier, but making each book a section seems like a flawed idea, because finding books in the config would be harder, you'd have to enumerate the keys. Instead, I think just a bookmarks section is reasonable, with file paths like: C:\Users\Quin\Dropbox\books\authors\Ami Spencer\Broken.epub = 1200,3029 for two bookmarks at those places.

danielw97:
Hi all,
As things stand, we now have a percentage saving mechanism using the config file.
It's still early days for this feature, so if anyone runs into issues or has feedback, please let us know.
This addition felt like a natural next step after adding position printing in the status bar.
Even at this stage, Paperback will likely become my daily EPUB reader due to how effective and lightweight it is.
Thanks again to @trypsynth for the solid groundwork.

aryanchoudharypro:
so will it open the book if i reopen the app after closeing it? 

[...]

PR #15, opened 2025-07-19
Refactor: remove duplicate position saving logic

danielw97:
Just a quick refactoring commit this time. as I mentioned in #11 I was observing a significant slowdown on a system with low resources when closing paperback.
I did some investigation, and found several instances where the position saving was being duplicated when the app was closing, i.e. not only was the position save timer still running, but it was also saving positions in multiple places which didn't help.
The biggest speedup was stopping the timer earlier, and has reduced the close time of the app by half.
Let me know your thoughts as always, I'm also happy to hold off and do a bigger refactoring commit later on if more issues are found or if you want any more changes.

### DIFF
--- a/src/app.cpp
+++ b/src/app.cpp
@@ -19,21 +19,6 @@ bool app::OnInit() {
 }
 
 int app::OnExit() {
-	if (frame) {
-		auto* notebook = frame->get_notebook();
-		for (size_t i = 0; i < notebook->GetPageCount(); ++i) {
-			auto* page = notebook->GetPage(i);
-			auto* data = static_cast<user_data*>(page->GetClientObject());
-			if (data && data->textbox) {
-				long position = data->textbox->GetInsertionPoint();
-				wxConfigBase* config = wxConfigBase::Get();
-				if (config) {
-					config->SetPath("/documents");
-					config->Write(data->file_path, position);
-				}
-			}
-		}
-	}
 	if (conf) conf->Flush();
 	return wxApp::OnExit();
 }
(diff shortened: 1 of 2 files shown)


trypsynth:
This all looks good, but it seems you ran clang-format over the code, which fucked up some formatting unfortunately. This is entirely my bad, the contributing guide tells people to do this and I wanted to use it to make my life easier, but I do nlooooot appreciate it mixing tabs and spaces in the list of menu IDs for example, nor do I appreciate it putting that ternary across two lines. I can push a commit to fix the formatting or you can, up to you. Either way I'm now pissed off and going to dig heavily through clang-tidy's docs to see how to turn this bullshit off. It annoyed me when I ran into it, but it's even more annoying when it makes me not want to merge an extremely solid seeming PR just because of its bullshit formatting.

danielw97:
sorry for that, that's also my bad for running it at the last minute.
I should be able to fix this now, hopefully it will merge cleanly and I'll refrain from using it for the moment.
I'm trying to be more conscious of code style, which is the main reason I ran it at the end.

trypsynth:
oh shit, seems my formatting the code conflicted with you doing it, because git is...well...git. I'm super sorry about that, can you resolve the conflicts or would you like me to do it?

danielw97:
Sorry for our commits colliding like that, it hopefully shouldn't be an issue in future. Would you mind resolving that conflict? I'm not able to see it in the cli when working with my remote, and the git web editor's not behaving the way I would expect it to on this occasion.
Thanks for fixing the .clang-format as well. I could also just reopen the pr if that's faster, as it's merging from a different branch, so I could just run the .klangformat after merging it.

danielw97:
thanks a lot, appreciate the fast merge. ironically I think I figured out how to fix that conflict in the cli right after my last message, but thanks for tackling it! least I'll know for next time.

PR #87, opened 2025-10-17
links support

### LINKED ISSUE #74: Support for Links in EPUB and DAISY 2 (and Other Linkable Formats)
It would be great if Paperback could support hyperlinks inside EPUB files, and later, inside DAISY textbooks once DAISY support becomes available. Many other formats also include internal or external links, and supporting them would significantly improve the reading experience.
Links are an essential part of many modern eBooks, especially EPUB files. They allow readers to quickly navigate from one section of the book to another, for example, from a reference in a chapter to a footnote, an appendix, or a glossary entry. Without working link support, these cross-references lose their purpose and make the reading flow less intuitive. Supporting links ensures a more natural and interactive reading experience, similar to how navigation works in web browsers or e-reading apps like QRead and Bookworm.
Suggested Implementation:
L: Move to the next link.
Shift + L: Move to the previous link.
Enter: Activate the current link.
Go to Link Dialog: A simple “Go to link…” dialog box could allow users to manually type or select a specific link destination within the book.
Once DAISY is supported, link handling could also be extended to that format, as well as any other text-based formats containing hyperlink or cross-reference structures.
Adding this functionality would make Paperback even more versatile and accessible for users who rely on internal navigation within eBooks.

aryanchoudharypro:
Support for Links in various formats. 
closes #74

### DIFF
--- a/src/document_manager.cpp
+++ b/src/document_manager.cpp
@@ -340,7 +340,7 @@ void document_manager::go_to_previous_link() {
 	const marker* link_marker = doc->buffer.get_marker(prev_index);
 	if (link_marker) {
 		go_to_position(link_marker->pos);
-		speak(_("Link: " + link_marker->text));
+		speak(_("Link: ") + link_marker->text);
 	}
 }
 


aryanchoudharypro:
shouldn't be hard to integrate because we already have word links in, but i have nothing to test. this

PR #98, opened 2025-10-19
initial table support

### LINKED ISSUE #81: Support for displaying tables in books or other files that contain tabular data
Currently, Paperback does not support tables. Tables appear in many books, and adding support for them would make Paperback more suitable for educational environments.
For example, QRead supports tables. When a user reaches a table in QRead or presses T/ SHIFT+T, they can even press the Enter key, and the table is displayed in a wxWebView with full support for table navigation keys in screen readers. However, I am not sure whether rendering through wxWebView is even possible in Paperback, or how compatible such an approach would be with security principles.

aryanchoudharypro:
this pr adds table support on books using html or xhtml 
closes #81

### DIFF
--- a/lib/src/html_to_text.rs
+++ b/lib/src/html_to_text.rs
@@ -234,7 +234,6 @@ impl HtmlToText {
 		self.finalize_current_line();
 		let table_html = Self::serialize_node(node, document);
 		let mut placeholder_text = "table: ".to_string();
-
 		let mut caption = None;
 		for child in node.children() {
 			if let Node::Element(e) = child.value() {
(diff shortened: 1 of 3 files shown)


trypsynth:
Ooh, I didn't see that there were any. I'm busy today and tomorrow until the evening but if you haven't done it by then I may look at it. I plan on making this land in 0.7 instead of 0.6, but it's easy to resolve conflicts early on.

trypsynth:
Awesome, thanks. I'll do some researching on that later this week and let you know what I find.

PR #106, opened 2025-10-20
added odp parser

### LINKED ISSUE #105: Add ODP support
We should support the open document presentation format, much like how we currently support the open document text format and PPTX files. Whether this should be put in the odt_parser and it should be renamed to open_document_parser or similar or put in its own file is left as an excersize for the implementer, whatever they think is best.

aryanchoudharypro:
added opd parser, is able to extract text and behaves just like the pptx parser. 
links not tested. 
closes #105


trypsynth:
Damn, that was fast! I'll see if I can find one with links in the next day or so, but then I'll review and merge this! Thanks!

aryanchoudharypro:
this update is going to be huge!

Issue #30, opened 2025-08-04
Support FB2 documents.


trypsynth:

trypsynth:
Yep, Duckduckgo's great for finding these. https://filesamples.com/index.php/formats/fb2

aryanchoudharypro:
thanks, i guess this is also coming if i don't sleep! 
else later

aryanchoudharypro:
@trypsynth this didn't get closed.

PR #107, opened 2025-10-21
fb2 parser

aryanchoudharypro:
closes issue #30

### DIFF
--- a/src/fb2_parser.cpp
+++ b/src/fb2_parser.cpp
@@ -26,25 +26,10 @@
 #include <sstream>
 #include <wx/log.h>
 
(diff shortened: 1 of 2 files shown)


aryanchoudharypro:
> Ah, makes sense. Debug code that slipped through to prod. NP at all, and again thanks for the contribution, please don't take my criticism personally, you just have to have fairly high code standards when you run a big open source project. One other thing, this isn't a big deal and I can just fix it myself as I do, but in a lot of your PR's, when you modify files, you delete the blank line from the end. The stylistic convention in Paperback's code is to keep it. Once again this is incredibly minor, but it does make your diffs slightly bigger originally and I go through and add them myself.

yeah i get it, just learning as i go. I will not remove the end line from now on.

PR #108, opened 2025-10-21
fixed the parsing of some markdown documents

aryanchoudharypro:


trypsynth:
Thanks again for the contribution, but I just pushed a simpler fix to master that works here, can you confirm?

PR #116, opened 2025-10-23
added a navigation wrap feature, off by Default

aryanchoudharypro:
this pr adds a navigation wrap feature that can be enabled in settings. When enabled, it will wrap the user to the end or start of the document, depending on the direction they are moveing.


trypsynth:
No, I got that, sorry. My question was Why only in the next* functions?

aryanchoudharypro:
because size_t can handle that number even if it gets large. i can change them to long as well if you want though.

PR #118, opened 2025-10-23
added sleep timer feature!

### LINKED ISSUE #117: Add a sleep timer
I'd like Paperback to be able to stop reading if I'm using say all in my screen reader.
My ideal implementation would be:

* You bring up the sleep timer from the tools menu and it asks for a number of minutes, preset to the last one you entered. Hitting OK here simply closes the app after that time has elapsed, without popup or noise or warning.
*  If there's already a timer running, pressing the option in the tools menu cancels it. The status bar could show "sleep" if it's set.
*  Bonus for me would be a restart on press option, so if I set the timer to 30 mins but pause reading to check something, I'm guaranteed another 30 mins of listening. I don't know how you could measure Human interaction versus screen reader scrolling, though.

aryanchoudharypro:
closes #117

### DIFF
--- a/src/config_manager.hpp
+++ b/src/config_manager.hpp
@@ -67,11 +67,9 @@ public:
 	config_manager& operator=(const config_manager&) = delete;
 	config_manager(config_manager&&) = default;
 	config_manager& operator=(config_manager&&) = default;
-
 	bool initialize();
 	void flush();
 	void shutdown();
-
 	wxString get_string(const wxString& key, const wxString& default_value = "") const;
 	bool get_bool(const wxString& key, bool default_value = false) const;
 	int get_int(const wxString& key, int default_value = 0) const;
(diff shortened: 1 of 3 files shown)


trypsynth:
@aryanchoudharypro sorry about this, but I think I just broke all your PRs again. The result is a much nicer config manager, but I apologise for that.

trypsynth:
Very nice, thanks! Couple tiny nitpicks:
* In the config manager, you removed a comment, added some unnecessary blank lines, and added a SetPath call. Please remove these, the SetPath call could be actually useful but the rest of the changes other than the ones that actually pertain to this PR should be removed.
* I appreciate you updating the typo at the top of main_window.cpp, but please undo the lowercasing of the word as in a couple of the comments at the top of these files.

Thanks again!

PR #123, opened 2025-10-25
added custom search in toc dialog so that space doesn't press the ok button and it is possible to have space when using single letter search

### LINKED ISSUE #121: Not urgent, allowing multiple word entry in TOC.
This is a pure nice to have, and should only be done if it's easy, in my view. When I open a file with a TOC in it, some of those entries may have spaces in them. If I want to go to a specific entry with a space in its title, I can't enter it because space activates the OK button, even when one is on the list of items. The best way to explain this is through an example, I think. To reproduce:

1. Download 

https://gutenberg.org/ebooks/8190.epub3.images

and open it in PB.

2. Go to the toc with CTRL+T. 
3. Now, you want to go to a story called "THE SECRET PLEASURES OF REGINALD"
4. If this was the first story in the book starting with the, you could just type in the and get to that story. You can now type in the but will find, after hitting t h e and space, that you are removed from the TOC and put at the start of the first story starting with "the", which is "THE NEW ADVERTISING".

I would like to type "the secr" and be put at, while still in the TOC, "THE SECRET PLEASURES OF REGINALD". Obviously, this isn't for this sort of book, I'm using it as an example, but it would be useful for complete collections, where I know what the title is but don't want to go through the list to find it. Again, this is a very minor thing, but I figured if it was easy to do with the current control, why not? People could still tab to OK and hit space, or just hit enter wherever they are.

aryanchoudharypro:
closes #121

### DIFF
--- a/src/dialogs.cpp
+++ b/src/dialogs.cpp
@@ -815,9 +815,7 @@ void toc_dialog::on_char_hook(wxKeyEvent& event) {
 	if (key_code >= WXK_SPACE && key_code < WXK_DELETE) {
 		search_timer_->StartOnce(500);
 		search_string_ += static_cast<wxChar>(event.GetUnicodeKey());
-		if (find_and_select_item_by_name(search_string_, tree->GetRootItem())) {
-			// Found and selected
-		} else {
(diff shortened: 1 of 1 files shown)


trypsynth:
Thanks!

PR #134, opened 2025-10-28
Add german translation

schulle4u:
This adds a German translation file. I changed the keyboard shortcuts for jumping between sections in my language file, since square brackets are rather inconvenient to use on German keyboard layouts. The German word for section is "Abschnitt", so I used the keys A and Shift+A for this purpose. Hope this won't break something in the future. 🙂

Steffen


trypsynth:
Thanks a ton for this!

PR #135, opened 2025-10-29
added search feature to all documents dialog

### LINKED ISSUE #133: Just a nice to have, search within recent document list dialog.
As almost always these days, this is just nice to have. Given that we are able to see opened documents, which is lovely, would it be possible to have a search box by file name in that dialog? Again, only if it's not too much work.

aryanchoudharypro:
closes #133


trypsynth:
Thanks for this, although it seems to have broken something in a really weird way. Escaping out of the dialog now causes a nice lovely crash! I love C++ </sarcasm>

PR #147, opened 2025-11-02
Add per-book export and import system for bookmarks and notes

### LINKED ISSUE #146: Add per-book export and import system for bookmarks and notes (.paperback files)
Inspired by the discussions here #131 
Currently, Paperback stores all bookmarks and notes in a single Paperback.ini file. While this works, it makes sharing or backing up data for individual books difficult and causes the master file to grow over time.
A more flexible approach would be to allow per-book export and import of bookmarks and notes, using a small .paperback file associated with each book. For example:
• When opening a book, Paperback could automatically look for a file named
filename.extension.paperback in the same directory as the book.
If found, it would import that data into the main configuration.
• Users could export their notes and bookmarks for a specific book into a .paperback file to share or back up.
• A “Forget this book” option could remove its data from the master file if desired.

All of these actions (import, export, forget) could conveniently be placed in the Document Info dialog, avoiding the need for additional menus or complexity.
This feature would make it much easier to exchange annotations between users, manage notes for study or collaboration, and keep the main configuration lightweight and organized.

aryanchoudharypro:
closes #146

### DIFF
--- a/src/main_window.cpp
+++ b/src/main_window.cpp
@@ -293,7 +293,6 @@ void main_window::update_ui() {
 	constexpr int doc_items[] = {
 		wxID_CLOSE,
 		wxID_CLOSE_ALL,
-		ID_EXPORT_TO_TEXT,
 		wxID_FIND,
 		ID_FIND_NEXT,
 		ID_FIND_PREVIOUS,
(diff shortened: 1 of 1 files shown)


amirsol81:
@aryanchoudharypro Thanks for your excellent work with this!

PR #197, opened 2025-11-17
Add Linux support with Nix packaging, Flatpak, desktop integration, and CI

ndarilek:
This gets most of Paperback running under Linux, either directly on Nix via `nix run`, or via Flatpak standalone bundles attached to each PR/release.

## What doesn't work?

* The ToC tree view is entirely inaccessible, I suspect due to WX's TreeView not being accessible under GTK3. It is still keyboard-navigable, but the widget isn't presented nor are any events related to it.
* Individual keyboard commands like "h" in the  document area output a terminal bell and do nothing. My suspicion is that GTK read-only TextViews block anything that looks like text input, that or either WX adds something.

Even with these limitations, having something I can throw just about any document I might want to read at under Linux and get back an accessible version is something I was going to write myself 6 months ago, so thanks for saving me the trouble! I think this is very useful even in its current form.

## Followup

Additional tasks to pursue once this gets merged--I'm happy to look into them but I'll put them here in case others might be interested:

* Other packaging options--AppImage at minimum. I'm prioritizing non-distro-specific formats for easier maintenance, though I needed Nix for the development setup so am willing to own that going forward.
* Publishing on [Flathub](https://flathub.org)
* Investigating whether WX's TreeView can be made accessible under Linux.
* Adding ctrl to all hotkeys on Linux in the document view to see if that makes them work.

## Disclosure

This PR was made largely with the help of Claude Code. As per my usual agentic workflow, I've reviewed every line of this PR. While nothing appears to `rm -rf /` or exfiltrate the nuclear launch codes, I can't speak to:
* The quality of the CMake additions
* The quality of the Flatpak build manifest. It looks good, albeit verbose, but Flatpak manifests kind of ar. I've done Flatpak manifests, but long enough ago that I don't know what current best practices are. But I've confirmed that the Flatpak installs and runs.
* The C++ changes. `-Werror` flagged a bunch of issues with my version of GCC, and not having done C++ in a couple decades, I let Claude fix them. At first glance it looks like pedantic style/correctness fixes, and a variable rename to prevent a shadowing-related error, but that's the smallest part of the diff so hopefully should be easy to review.
* The extracted icons. I needed icons for the .desktop files and Flatpak. Claude claims it extracted and resized the icons from the Windows installer. The description it gives me for the icons it added seems to make sense, but we should probably have someone visually confirm them (or confirm them via your own/someone else's VLLM analysis.)


trypsynth:
Thanks a ton for this! This is a fairly sizable diff and I do have software engineering employment, so my apologies if this takes me a little bit to merge, but thanks a ton for all your work! It does look like CMakeLists has some conflicts, but those should be easy to fix.
FWIW, I have the very same agentic coding workflow. Write a massive prompt, let Claude or Codex cook, review every single line multiple times.

ndarilek:
Please hold off just a bit, there's an odd issue where it segfaults if I attempt to open a book when launching from the desktop but works fine when opened from the CLI. Going to attach a debugger and investigate that today, should hopefully have a fix later this afternoon.

ndarilek:
Got it.

The good news is that it is accessible. The bad news is 2-fold:
1. There's some weird ordering issue I can't work out, where it seems like tree items are unordered even though they appear correct in both debugging output and Orca flat review. Arrowing up/down presented chapters/subsections in weird and random orders.
2. WXDataViewTreeControl reliably froze Orca whenever I attempted flat review to verify the contents were in fact ordered. I have to assume they were correct in the data model since the behavior persisted across both attempts, but I wasn't able to verify because of the consistent freeze.

So we're back to the weird hacky list, and someone smarter than me will have to figure it out. But as of now both the ToC and elements list use this hacky tree and are accessible.

Marking this as draft because I haven't fully reviewed the code myself yet. I'll do that tomorrow morning or early afternoon and undraft.

trypsynth:
@ndarilek finally getting around to reviewing this, thanks for all your work!

PR #223, opened 2025-12-30
Update Bosnian translation

tareh7z:
Updated Bosnian translation.

### DIFF
--- a/app/document_data.hpp
+++ b/app/document_data.hpp
@@ -40,6 +40,26 @@ struct toc_item {
 	toc_item& operator=(toc_item&&) = default;
 };
 
+struct session_document {
+	rust::Box<DocumentSession> session;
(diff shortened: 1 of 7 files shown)


hdzrvcc0X74:
I’m honestly not sure. I might be doing something wrong when it comes to syncing my fork with your repository and pulling those changes locally.

trypsynth:
How are you doing it? The way I would recommend doing so, if you're at least a tiny bit comfortable with the command line, is to grab the GitHub CLI, install and authenticate it, and then clone paperback from me. After that, each time you want to make a new translation update, make a new branch, add, commit, gh pr create, it'll take you through a little wizard, letting you push your branch directly to your fork based on my latest master, and then open the pr for you automagically.

PR #326, opened 2026-02-11
Add Japanese translation

kitabatake1013:
I have created a Japanese translation file based on the current `pot` file.

All existing messages have been translated, and the translation is ready for use. Future refinements may be made as needed, but the current version fully covers the existing strings.

I used Poedit to edit the `.po` file, so Poedit-specific metadata has been automatically added to the file header. If you would prefer not to include that metadata, I can manually remove it—just let me know.

Thank you for your review.


trypsynth:
Looks good to me, thanks! Poedit is what most of my contributors use :)

PR #403, opened 2026-03-26
Add maximum line length option for screenreader accessibility

vcaparica:
## Summary

- Adds a **Maximum line length** setting to the Options dialog (Reading tab), allowing users to define how many characters per line before wrapping at word boundaries
- Default is 0 (entire line, preserving current behavior). Setting it to e.g. 100 breaks long paragraphs into ~100-character lines so screenreaders read manageable chunks instead of entire paragraphs at once
- The wrapping algorithm replaces spaces with newlines at word boundaries, preserving exact character count — so bookmarks, cursor positions, find results, and TOC navigation remain accurate with no position mapping needed
- Also fixes an accessibility issue where checkboxes in the Options dialog were not announcing their labels to screenreaders, by setting explicit accessible names on all controls

## Technical Details

- `wrap_content()` in `text.rs`: processes each paragraph independently, splitting at word boundaries. Words longer than the limit are kept intact (no mid-word truncation)
- `apply_line_wrapping()` in `document.rs`: applies wrapping to the document buffer and rebuilds newline position tracking
- Wrapping is applied after parsing in `DocumentSession::new()`, before the document handle is created
- When the setting changes, all open documents are saved, closed, and re-opened with the new wrapping applied
- Includes unit tests for the wrapping function

## Test plan

- [ ] Open an EPUB with long paragraphs, verify lines display as full paragraphs with default setting (0)
- [ ] Set maximum line length to 100, verify paragraphs wrap at ~100 characters at word boundaries
- [ ] Verify screenreader reads one wrapped line at a time instead of entire paragraphs
- [ ] Verify bookmarks, find, and TOC navigation work correctly after changing the setting
- [ ] Verify changing the setting re-opens documents and restores cursor position
- [ ] Verify all Options dialog controls announce their labels to screenreaders

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -773,6 +773,7 @@ impl ConfigManager {
 			("navigation_wrap", false, None, None),
 			("check_for_updates_on_startup", true, None, None),
 			("recent_documents_to_show", false, Some(DEFAULT_RECENT_DOCUMENTS_TO_SHOW), None),
+			("max_line_length", false, Some(0), None),
 			("sleep_timer_duration", false, Some(30), None),
 			("language", false, None, Some("")),
 			("active_document", false, None, Some("")),
--- a/src/document.rs
+++ b/src/document.rs
@@ -171,6 +171,17 @@ impl DocumentBuffer {
 	pub fn newline_positions(&self) -> &[usize] {
 		&self.newline_char_positions
 	}
+
+	pub fn apply_line_wrapping(&mut self, max_width: usize) {
+		use crate::text::wrap_content;
+		self.content = wrap_content(&self.content, max_width);
+		self.newline_char_positions.clear();
+		for (i, c) in self.content.chars().enumerate() {
+			if c == '\n' {
+				self.newline_char_positions.push(i);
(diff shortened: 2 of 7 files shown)


amirsol81:
@trypsynth This can come in handy in cases where large paragraphs in Epub or DAISY books are presented on a single long line. This is oftentimes the case, and I myself sometimes struggle to move among sentences or phrases inside such paragraphs on a single line.

trypsynth:
Any thoughts on this? I've yet to get any justification for it and I've just confirmed that word wrap works perfectly in the current version minus a minor bug where it resets your cursor position in your focused document, which I'm now fixing.

amirsol81:
@trypsynth I can confirm that word wrap functions properly with the latest dev build.

trypsynth:
I've not heard from the author since they opened this PR, and users plus myself have confirmed word wrap works and does pretty much the same thing. The way this PR was written leads me to think it was entirely AI generated, and never checked. I am not against agentic coding; I use claude all the time, but I think this one can be closed.

PR #423, opened 2026-04-24
add hashing for saveing document data, fixed current tab not getting focused on session restore

### LINKED ISSUE #420: Don't use the file name and path for bookmarks and stored positions
Paperback currently uses the file path and name to associate bookmarks and saved reading positions. If the filename changes, the current progress is lost.
Could a checksum (MD5 or SHA) be used instead? This would preserve bookmarks and position even when reorganizing the library.

aryanchoudharypro:
closes #420

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -275,6 +275,8 @@ struct ConfigData {
 	find_history: Vec<String>,
 	#[serde(default)]
 	documents: HashMap<String, DocumentConfig>,
+	#[serde(default)]
+	path_hashes: HashMap<String, String>,
 }
 
 impl Default for ConfigData {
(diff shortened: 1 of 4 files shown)


trypsynth:
Nice work on this! It definitely solves the rename issue, but I'm a bit worried about it chugging on the first run. Since the migration hashes everything at once in the config loader, anyone with a big library is going to hit a massive hang on first startup. Also, if a file is missing, it falls back to hashing the path string. That means if I move a file while the app is closed, the IDs won't match and the bookmarks are still lost anyway. It'd be safer to compute these hashes lazily and store them permanently in the config so you aren't hitting the disk every time doc_key is called. Also, heads up that you're using Base64 in one spot and Hex in another; probably worth picking one and sticking to it!

PR #444, opened 2026-05-08
updated the document Readability format options to use the native WX Widgets functions, accept letter spaseing.

aryanchoudharypro:


trypsynth:
Hmm, this seems to have undone the logic that makes this logic not apply if readability settings are at their defaults. Worm now takes upwards of 10 seconds to load. Not sure why at first glance though.

trypsynth:
I was wrong actually, I had settings set to nondefault values and what I thought was it being slow with defaults was actually it being faster than before with configured settings. Well done!

PR #527, opened 2026-06-18
Don't show window until file is loaded

Keithcat1:
For large files, showing the window while loading can cause NVDA to become unresponsive for several seconds.

### DIFF
--- a/crates/paperback/src/ui/app.rs
+++ b/crates/paperback/src/ui/app.rs
@@ -52,6 +52,7 @@ impl PaperbackApp {
 		MAIN_WINDOW_PTR.store(Rc::as_ptr(&main_window) as usize, Ordering::SeqCst);
 		wxdragon::app::set_top_window(main_window.frame());
 		let pipe_server = start_pipe_server(&Rc::clone(&main_window));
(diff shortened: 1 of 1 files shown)


trypsynth:
Thanks for this, although I'm sorry to say this is not the approach I had in mind. This is why I like people to open issues before making changes like this, but we can always move over to a discussion if necessary. My ideal UX would be a loading dialog that shows up for only like 200 ms if yo uload a teeny book, a few seconds if you load a giant one, or a tad longer if you're loading multiple documents. Importantly, it should not flash for every document opened, it shouldn't give you verbose speech, just something like "loading, dialog", etc.

Keithcat1:
HMM, I hadn't thought of that. The problem is that this has to be moved off of the main thread as far as I can tell, and probably should avoid instantiating widgets. It seems interlinked with the GUI code.

trypsynth:
Nods, it will have to be invoked on a background thread for sure. Are you interested in doing this, or should I close this pr and do it myself at some point?

Keithcat1:
I don't really have a use for it and kind of prefer my proposed behavior because it's quieter. I'll close.

PR #532, opened 2026-06-19
feat: add View Source to open a document's source in a new tab

LeonarddeR:
## Summary

Adds a browser-style **View Source** command (Ctrl+U, Tools menu) that opens the
underlying source of the current document in a new text tab, positioned at the
current reading location — mirroring "View Source" in a web browser.

## Behavior

- **HTML / XHTML**: re-reads the original file from disk.
- **EPUB**: extracts the HTML of the *current spine section* from the archive
  (browser-like — the "page" you're on), not the whole book.
- **Markdown**: shows the raw `.md`.
- **Unsupported (binary) formats** (PDF, DOCX, MOBI, ODT/ODP, PPT, etc.): show an
  error dialog, consistent with Open in Web View.

The source opens in a read-only text tab titled `Source: <name>`. The tab is
untracked (not added to recent documents, no saved position) and the temp file is
written to a stable per-document path under the system temp dir, so re-invoking
focuses the existing source tab rather than accumulating files.

## Position mapping

Reuses the Web View position-mapping logic:
- HTML/XHTML/EPUB: `XmlToText::find_anchor_byte_offset` maps the reading position
  to a source byte offset, converted to a caret offset.
- Markdown: the `pb-block-N` block anchors already recorded during parsing are
  mapped back to the raw source via a new `markdown::block_source_offset` helper
  (pulldown-cmark source offsets), landing the caret at the start of the current
  block/line — the same granularity the Web View uses.

## Tests

- `markdown::block_source_offset` unit test.
- `view_source` tests: HTML caret mapping, EPUB/Markdown caret, unsupported format
  returns `None`, `source_view_available` matrix.
- Full `paperback-core` suite: 327 passing. Both crates build.


trypsynth:
Hi, thanks for this! Can you please provide me some examples of real-world use cases before I go ahead and review/merge? This is never something I imagined Paperback doing, and it could be cool, but I'm trying to think of when soneone might want to use it over a browser.
Thanks as always for your contributions!

LeonarddeR:
Good point @trypsynth 

For me, this feature comes in handy especially when working with markdown. I tend to write in markdown a lot, and Paperback is a lovely way to read markdown rendered without actually needing to convert the document to another format (e.g. to docx or html) with pandoc.

That said, when making a mistake in markdown, the result can get clunky of course. Then, the view source command is a quick way to jump to the source of the document to see what goes wrong without having to open it in a side editor and find the right position in the document to fix it up.

I hope that's enough ground for the feature too you. If not, feel free to discard it.

trypsynth:
@LeonarddeR fair enough, I don't have an issue with the feature at all, just didn't know what one might use it for. This makes sense though, thanks! I'll review and merge this this weekend.

trypsynth (approved):

PR #541, opened 2026-06-27
feat(tables): render tables inline toggle (#535)

### LINKED ISSUE #535: UNiformize table handling
I can see at least four ways tables are handled in Paperback currently:

* In PDF's, `[Table]` is shown. Pressing enter on it opens the table in a web view
* In HTML/ open text documents and MarkDown, the table is shown one cell per line. Enter on the table also shows the table in a web view.
* In Word: `Table: <first row>` is shown, enter opens web view.
* In PowerPoint, tables are shown one cell per line but it looks like they aren't detected as tables.

What I'd like to propose, is a toggle in settings that either shows `[Table]` or renders the table in the document consistently. I also wonder whether the one cell per line approach is sufficient. Wouldn't it be better to show the table tab separated per row?

LeonarddeR:
## Summary

Closes #535: a "Render tables inline" toggle (Options, Readability tab) that
uniformly controls table rendering across all document formats.

- **ON (default):** the table rendered inline as tab-separated rows (cells separated by
  tab, rows by newline).
- **OFF:** uniform placeholder `[Table]: <first row>` (the table's first row, with tabs
  replaced by spaces; just `[Table]` when the table has no rows).

Both representations are produced by a single helper (`html_table_to_display`), so PDF,
Word, ODT, HTML/EPUB and Markdown render identically.

## Design

The flag is threaded from the config through `ParserContext` into each parser, which
emits the table text at parse time. Toggling the setting re-parses the open documents.
Markers get correct positions during the normal parse pass, so there is no post-parse
offset-shifting and no dual-variant cached document.

## Known limitations

- PowerPoint tables are still not detected; out of scope, follow-up.

### DIFF
--- a/crates/paperback-core/src/config.rs
+++ b/crates/paperback-core/src/config.rs
@@ -122,6 +122,8 @@ pub struct AppSettings {
 	pub restore_previous_documents: bool,
 	#[serde(default)]
 	pub word_wrap: bool,
+	#[serde(default = "default_true")]
+	pub render_tables_inline: bool,
 	#[serde(default)]
 	pub navigation_wrap: bool,
 	#[serde(default)]
@@ -174,6 +176,7 @@ impl Default for AppSettings {
 		Self {
 			restore_previous_documents: true,
 			word_wrap: false,
+			render_tables_inline: true,
 			navigation_wrap: false,
 			find_match_case: false,
 			find_whole_word: false,
@@ -412,6 +415,7 @@ impl ConfigManager {
 		match key {
 			"restore_previous_documents" => data.app.restore_previous_documents,
 			"word_wrap" => data.app.word_wrap,
+			"render_tables_inline" => data.app.render_tables_inline,
 			"navigation_wrap" => data.app.navigation_wrap,
 			"find_match_case" => data.app.find_match_case,
 			"find_whole_word" => data.app.find_whole_word,
@@ -459,6 +463,7 @@ impl ConfigManager {
 			match key {
 				"restore_previous_documents" => data.app.restore_previous_documents = value,
 				"word_wrap" => data.app.word_wrap = value,
+				"render_tables_inline" => data.app.render_tables_inline = value,
 				"navigation_wrap" => data.app.navigation_wrap = value,
 				"find_match_case" => data.app.find_match_case = value,
 				"find_whole_word" => data.app.find_whole_word = value,
@@ -1106,5 +1111,16 @@ mod tests {
 		let a = config.get_doc_key("book-a.epub");
 		let b = config.get_doc_key("book-b.epub");
 		assert_ne!(a, b);
+	}
+
+	#[test]
+	fn render_tables_inline_round_trips() {
+		let mut config = ConfigManager::new();
+		config.initialized = true;
+		assert!(config.get_app_bool("render_tables_inline", true));
+		config.set_app_bool("render_tables_inline", false);
+		assert!(!config.get_app_bool("render_tables_inline", true));
+		config.set_app_bool("render_tables_inline", true);
+		assert!(config.get_app_bool("render_tables_inline", true));
 	}
 }
--- a/crates/paperback-core/src/document.rs
+++ b/crates/paperback-core/src/document.rs
@@ -532,12 +532,15 @@ pub struct ParserContext {
 	pub file_path: String,
 	pub password: Option<String>,
 	pub forced_extension: Option<String>,
+	/// When `true`, parsers emit each table's full tab-separated rendering inline; when `false`,
+	/// they emit a `"[Table]: <first row>"` placeholder. Threaded into each parser at parse time.
+	pub render_tables_inline: bool,
 }
 
 impl ParserContext {
 	#[must_use]
 	pub const fn new(file_path: String) -> Self {
-		Self { file_path, password: None, forced_extension: None }
+		Self { file_path, password: None, forced_extension: None, render_tables_inline: true }
 	}
 
 	#[must_use]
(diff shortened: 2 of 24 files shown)


LeonarddeR:
@trypsynth I'm afraid it became a pretty huge diff.
This will also change default behavior for pdf, as I made inline the default as per https://github.com/trypsynth/paperback/issues/535#issuecomment-4814396284

trypsynth (commented):
## Code review

Good foundation — the shared `table_text.rs` helper gives uniform rendering across all formats, the migration from `marker.text.chars().count()` to `marker.length` is the right fix for the boundary check, and the test coverage is solid. A few issues to address:

---

### Bug: ODT drops XML anchor IDs inside table cells

`process_table` in `odt.rs` loses its `id_positions` parameter. The old code called `traverse_children(cell, buffer, id_positions)` which registered any `id` attribute found on elements nested inside table cells. The new code calls `collect_element_text(cell)` — a pure text extractor that never touches `id_positions`.

Result: any ODT element inside a table cell that carries an `id` attribute (bookmarks, footnote anchors, cross-reference targets) will be silently absent from `id_positions`. Internal links pointing to those anchors will fail to navigate.

Fix: after building `html_content`, call `traverse_children(cell, buffer_for_ids, id_positions)` on a scratch buffer to collect IDs — or pass `id_positions` back into `process_table` and register them at the stable display offset of the cell's start position.

---

### Bug: empty inline table in PDF adds a spurious blank line

In `append_pdf_table_to_buffer`, when `render_tables_inline = true` and the table has no rows, `html_table_to_display` returns `""`. Rust's `"".split('\n')` yields one item (`""`), not zero. The loop body runs once: it pushes `(pos, "")` into `current_lines_info` and appends `"\n"` to both `buffer` and `page_display_text`.

`HtmlToText`/`XmlToText` avoid this via the `display_lines_and_length("")` guard that returns `(Vec::new(), 0)`. The already-exported `display_lines_and_length` from `table_text.rs` could replace the raw `split('\n')` here:

```rust
let (lines, _) = crate::parser::table_text::display_lines_and_length(&display_text);
for line in lines { ... }
```

---

### Nit: `"table: "` prefix in Word marker text is inconsistent

`process_table` in `word.rs` initialises `table_caption` as `String::from("table: ")` and appends cell content to it, so `Marker::with_text` stores e.g. `"table: Kop 𝄞"`. Every other parser (ODT, HTML, XML via `table_render_bundle`) stores just the first-row content without a prefix. Since `table_caption_from_html` is now available in `table_text.rs`, the caption-building loop in `process_table` can be replaced with a call to it, which would give consistent labels across formats.

---

### Nit: PR description says "OFF (default)" but the code defaults to ON

`#[serde(default = "default_true")]` and `Default::default()` both set `render_tables_inline = true`. Assuming that's intentional (inline rendering on by default seems right for most users), the PR description's "OFF (default)" bullet is just wrong — worth updating before merge so the description doesn't confuse future readers of the changelog.

---

### Cleanup: `table_caption_from_html` is exported but has no callers

`pub fn table_caption_from_html` in `table_text.rs` is never called from any parser — `HtmlToText`/`XmlToText` use `table_render_bundle`, `word.rs` builds its own caption, and `pdf.rs` uses `html_table_to_display`. As a `pub` item it becomes API surface; a future author adding a new parser will reach for it over `table_render_bundle` and pay an extra HTML parse for free. Either make it `fn` (private) or have the Word parser adopt it (which would also fix the prefix inconsistency above).

---

### Cleanup: `push_finalized_line` copy-pasted verbatim in two structs

The three-line method body is identical in `HtmlToText` and `XmlToText`. A future change to the `+1` newline accounting applied to one but not the other would silently diverge position tracking between HTML and XML documents. A shared trait or a standalone `fn push_finalized_line(lines: &mut Vec<String>, cached_len: &mut usize, line: String)` would prevent that.

LeonarddeR:
Thanks for the review — all six addressed:

**1. ODT drops IDs inside table cells** — `process_table` now takes `id_positions` and walks the table's descendants, registering every in-cell `id` at the table's start position. I went with table-start rather than the scratch-buffer route: once a table collapses to a placeholder/TSV the cells have no individual display offset, and a scratch buffer would record scratch-relative offsets. Registering each anchor at the table start lands the reader on the table, which is the only stable target. New test covers OFF and ON.

**2. Empty inline PDF table blank line** — `append_pdf_table_to_buffer` now uses `display_lines_and_length`, which returns no lines for empty input instead of one `""`. Empty inline table → nothing appended, zero-length marker. New test added.

**3 + 5. Word `"table: "` prefix / unused `table_caption_from_html`** — fixed together: `process_table` in `word.rs` drops the manual caption loop and derives the caption via `table_caption_from_html(&html_content).unwrap_or_else(|| "table".into())`, matching the HTML/XML default. Prefix gone, labels consistent, and the exported fn now has a caller. Test expectation updated.

**4. "OFF (default)" in the description** — updated the PR body; ON is now listed as the default.

**6. Duplicated `push_finalized_line`** — extracted to a shared `table_text::push_finalized_line(lines, cached_len, line)`; both `HtmlToText` and `XmlToText` delegate to it, so the `+1` newline accounting can't diverge.

`cargo test -p paperback-core` green (365 passed).

trypsynth (approved):
All six items addressed and verified against the updated diff:

- **ODT IDs**: `process_table` takes `id_positions` again; the upfront `descendants()` walk registers every in-cell `id` at the table start, and the new `odt_table_cell_id_registered_at_table_start` test pins the behaviour in both modes.
- **PDF empty table**: `append_pdf_table_to_buffer` now routes through `display_lines_and_length`, which guards the empty case — new test confirms `buffer.content == ""` and `lines_info.is_empty()`.
- **Word caption prefix + dead `table_caption_from_html`**: fixed together — `process_table` in `word.rs` now calls `table_caption_from_html`, giving consistent first-row labels across formats with no `"table: "` prefix, and the exported function now has a caller.
- **PR description**: updated to ON-as-default.
- **`push_finalized_line` duplication**: extracted to `table_text::push_finalized_line`; both converters delegate to it.

LGTM.

PR #551, opened 2026-06-30
Reorder right-to-left text in PDF extraction to logical order

LeonarddeR:
## Problem

pdfium returns characters in the order they're drawn on the page, which for right-to-left languages (Hebrew, Arabic, …) is often the visual left-to-right order. The result is that words come out reversed and accent marks get separated from their letters. Other PDF engines reorder this to logical reading order; pdfium does not. Since the GUI and screen readers read the same text buffer, RTL PDFs are garbled for everyone.

## Fix

New module `crates/paperback-core/src/parser/util/bidi.rs` with `reorder_line()`, wired into both PDF text-extraction paths in `pdf.rs`. It uses each character's x position on the page to rebuild logical order:

- It only touches runs that are actually stored reversed. Text already in correct order is returned unchanged, so well-formed documents never regress.
- For reversed runs it re-attaches accent marks to the correct letter, reverses the run, and mirrors brackets as needed.
- Accent marks are re-attached by x position: a mark is drawn at the same x as the letter it sits on (no horizontal advance), so even when pdfium emits it out of sequence it's matched to the nearest letter by x.
- Character classification and bracket mirroring come from `icu_properties`, so Hebrew, Arabic, and other RTL scripts are all handled.

## Testing

- 8 new unit tests (Latin unchanged, Hebrew and Arabic reversed→logical, already-correct text untouched, mark re-attachment, Latin word inside an RTL line).
- Full `paperback-core` suite passes (378 tests).
- Verified with `pb`: a Hebrew PDF already in correct order extracts identically (no regression); a mixed Dutch/Hebrew PDF's reversed words now read correctly.

## Known limitations

- Some PDFs duplicate or misplace accent marks at creation time — every extractor reproduces this, so it can't be fixed here.
- On reordered lines, stray spaces and bracket placement at language boundaries can be slightly off, but it's a clear improvement over the fully-reversed originals.


trypsynth:
Thanks for this, it seems incredibly well thought out. The fast-path logic to prevent regressions on already-logical PDFs and the proximity matching for floating diacritics show a deep understanding of PDF extraction pain points that I've had to familiarize myself with too, very well done. Just a couple minor things before approval:
1. CodePointMapData::<BidiClass>::new() is being called per-line inside reorder_line. Since reorder_line is invoked inside a loop over all lines in the document, this is going to probably heavily impact performance. Can we instantiate this once at the parser/page level and pass it down as a reference?
2. Is the 0.5 threshold in has_visual_rtl_run pulled from a specific standard or derived empirically from your test documents? (Just want to make sure tightly kerned fonts won't accidentally trigger it).

LeonarddeR:
Thanks for your feedback.

> CodePointMapData::<BidiClass>::new() is being called per-line inside reorder_line. Since reorder_line is invoked inside a loop over all lines in the document, this is going to probably heavily impact performance. Can we instantiate this once at the parser/page level and pass it down as a reference?

I checked the `icu_properties` source. That constructor is a `const fn`:

```rust
pub const fn new() -> CodePointMapDataBorrowed<'static, T> {
    CodePointMapDataBorrowed { map: T::SINGLETON }
}
```

`T::SINGLETON` is a `&'static` reference to the trie data, which is baked into the binary at compile time via the `compiled_data` feature.

> Is the 0.5 threshold in has_visual_rtl_run pulled from a specific standard or derived empirically from your test documents? (Just want to make sure tightly kerned fonts won't accidentally trigger it.)

It's empirical — tuned against the test PDFs, not derived from any PDF or Unicode standard.

On the kerning concern specifically: the comparison only trips when the x-position *increases* between two consecutive right-to-left base characters (`x > px + 0.5`). Tighter kerning shrinks that delta, which moves it further from the threshold, not closer. So the failure mode from tight kerning would be a missed reorder (a visually-stored RTL run that doesn't get flagged), never a spurious one (a logically-stored run getting incorrectly flagged as needing reordering).

The scenario that could actually cause a miss is the opposite one: a PDF that renders its text at a very small scale (via the CTM) could compress the real position deltas of a genuinely visually-stored RTL run to below 0.5, causing it to be left unreordered. That's already the "default, unscaled CTM" assumption called out in the doc comment above the threshold — a known limitation rather than a new one raised by this question.

trypsynth:
Okay, fair enough. LGTM, thanks for this, and good work!

PR #554, opened 2026-07-01
Add property page accessible role for settings tabs and anounce document name on tab switch

LeonarddeR:
- Add a small `accessibility` module exposing `AccProps`/`AccessibleExt` for setting accessible name/role on any wxdragon widget
- Set `ROLE_PROPERTYPAGE` + the tab's title as accessible name on the three Options dialog tab panels, so screen readers announce them as pages on Ctrl+Tab
- Set the accessible name on each document tab's rich edit control to the document title. This ensures that switching between them with ctrl+tab announces the newly selected document properly.
Note that this only works on Windows, I'm too unfamiliar with Mac to see what might need changing there.
Also this slightly increases verbosity on the edit control because it mentions the title.

### DIFF
--- a/crates/paperback/src/ui/dialogs/options.rs
+++ b/crates/paperback/src/ui/dialogs/options.rs
@@ -7,6 +7,8 @@ use std::{
 
 use paperback_core::config::{ConfigManager, HotkeyConfig, ReadabilityFont};
 use patois::{t, ui::populate_language_choice};
+#[cfg(target_os = "windows")]
+use wxdragon::accessible::AccRole;
 use wxdragon::prelude::*;
 
 use super::DIALOG_PADDING;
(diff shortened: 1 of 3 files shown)


trypsynth:
@LeonarddeR Thanks for this! Some thoughts:

> Add a small accessibility module exposing AccProps/AccessibleExt for setting accessible name/role on any wxdragon widget

Can you please try upstreaming this into wxDragon instead? I did this for macOS after I confirmed my implementation of labeling combo boxes and it got accepted. Slightly more work, but worth it in the long run IMO

> Also this slightly increases verbosity on the edit control because it mentions the title.

I'll have to try this out to see how much it annoys me, I might end up voting for making that configurable or finding a different way to do it, but time will tell on that one.
Thanks for this as usual!

trypsynth:
Thanks for this! Would it be possible to make the property page for documents a toggle? I tend to read books with very long titles and this adds unnecessary verbosity IMO, especially when alt + tabbing back into the window, but I understand why someone would want it, and in the settings dialog it should always apply.  I'm thinking a announce document title on tab switch or something for the checkbox label, unchecked by default. As always thanks for all your great contributions!

LeonarddeR:
Agreed that it can be a bit long at times. But when document title as a label on the text control is problematic, it might be better to use the live region to announce documents switches instead. What do you think?

LeonarddeR:
I just added the live region way, feels more smooth to me. It required a change to wx dragon though, so lets wait for https://github.com/AllenDang/wxDragon/pull/161

trypsynth:
@LeonarddeR Thanks for this, the code looks good, but I and @aryanchoudharypro had the following two thoughts:

1. Can you please remove the announcements for closing documents, especially closed all and announcing what document just got closed? If you need to keep a live region announcement in the document close path to announce what document got focused that seems fair enough, but this adds way too much verbosity for our liking.
2. This relates to point number 1, but can the announcements please just be the document title? "Switched to" just adds extra unnecessary verbosity. This ties into the end of point number 1.


Thanks!

LeonarddeR:
@trypsynth Thanks, actually makes sense to me. I hope this fits you better.

PR #564, opened 2026-07-02
Fix unresponsive cancel and unguarded concurrent runs in the updater

tareh7z:
The update flow (via ship-shape) closed the progress dialog when Cancel was pressed, but never stopped the background download thread. The blocking read loop just kept pulling bytes and writing to disk until hitting its 10-minute timeout. On slow connections, a "cancelled" download would keep running invisibly in the background.

There was also no guard against concurrent update checks. A second check (like a manual click racing the silent startup check) would let two downloads race on the same temp file, allowing a stale, cancelled flow to close the progress dialog of a newer one.

Since ship-shape is an external crate, it is now vendored into crates/ship-shape and patched via [patch.crates-io]:

* Download_update_file: Now checks a cancellation flag before network round-trips and on every chunk of the read loop to abort immediately. It also cleans up partial temp files on cancellation or error.

* Run_update_check: Now returns early (no-op) if a check or download is already active, closing the race condition hole.

Both crates/ship-shape and the paperback binary build clean against the patched dependency.


trypsynth:
Thanks for this! Can we avoid vendoring ship-shape though? Just use a crates.io patch directive to point to your fork to confirm it works, and open a PR against ship shape too.

tareh7z:
Hi @trypsynth!

Done. Vendoring is gone and the patch now points at my fork with the fix which mirrors the [PR](https://github.com/trypsynth/ship-shape/pull/2) I've opened against     ship-shape itself, until it's merged upstream and released.

Thanks.

trypsynth:
Hi there, thanks! I've just published 0.1.5, please update to that, make sure it builds, and then I'll merge :)

tareh7z:
Hi @trypsynth!

Updated to ship-shape 0.1.5 and dropped the patch. Builds clean.

Thanks.

Issue #557, opened 2026-07-01
Preserve bold, itallic, underline


LeonarddeR:
Paperback currently discards bold/itallic/underline in the document reader. 
I think it is stuff that is easy to preserve in the text control while it also gives the advantage that this info won't get lost. Also it opens possibility to introduce text navigation, i.e. go to the next/previous bold/itallic/underline, go to the next differently formatted text, etc.

trypsynth:
Nods, I agree. Although at this rate, we might have to start making you able to re-assign your quick navigation keys if we start running out, haha! Good idea though.

aryanchoudharypro:
I'll first be adding support for just formatting the text, then we can see how we want to do next diffrently formatted bold or other elements.

LeonarddeR:
I let claude already work on a prototype over the night, but if you @aryanchoudharypro want to do it, feel free.

PR #568, opened 2026-07-03
Preserve font face and size for bold/italic/underline runs

LeonarddeR:
Follow-up / fixup for #557.

## Problem
On Windows, bold/italic/underline runs rendered in Arial 10pt instead of
the reader's font. wxMSW's `wxTextCtrl::SetStyle` rewrites the *entire*
font of a range whenever the `wxTextAttr` carries any font attribute — it
masks `CFM_FACE | CFM_SIZE | ...` unconditionally and fills unset fields
from a default font (Arial 10pt). Applying each marker as a
single-attribute `wxTextAttr` therefore reset the run's face/size, and
overlapping markers (e.g. a bold word inside an italic sentence) clobbered
each other's styles.

## Fix
Merge the (possibly overlapping) markers into non-overlapping segments
carrying the union of active styles, then apply one *full* font per
segment (current control font + the segment's styles). Preserves face/size
and composes overlapping styles correctly on every platform.

Added unit tests for the marker-merge logic.

## Verification
Built and opened a document with bold/italic/underline and combined
bold+italic runs — all render in the reader font at the correct size, and
the bold+italic run shows both styles.

### DIFF
--- a/crates/paperback/src/ui/document_manager.rs
+++ b/crates/paperback/src/ui/document_manager.rs
@@ -1041,29 +1041,121 @@ pub fn apply_readability_format_to_ctrl(
 	}
 }
 
+/// A non-overlapping run of text with the union of bold/italic/underline
+/// styles active over it.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct FormatSegment {
+	pub start: i64,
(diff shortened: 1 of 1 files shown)


LeonarddeR:
Cc @aryanchoudharypro

PR #575, opened 2026-07-06
Bundle wxWidgets translations into the binary

### LINKED ISSUE #570: Wx strings not translated
The translatable strings in wx don't work in Paperback. This is especially noticeable in the update progress dialog.

This relies on https://github.com/AllenDang/wxDragon/pull/159. When that's done, the wx translations can be bundled in the binary and a translation loader can be implemented to load them.

LeonarddeR:
Closes #570

## Problem

wxWidgets' own UI strings (progress-dialog labels, "Cancel", "Save", etc.) were never translated in Paperback. wx loads its standard `wxstd` catalog from a `share/locale` tree on disk, which doesn't exist next to an installed Paperback executable — so those strings always fell back to English regardless of app language.

## Approach

Bundle wx's `wxstd` `.mo` catalogs into the binary at compile time and serve them to wxWidgets at runtime, mirroring how [`patois::embed_domain!`](https://docs.rs/patois) already bundles Paperback's own translations.

### Changes

- **`build.rs`** — `embed_wx_translations()` scans wxWidgets' catalogs (emitted by `wxdragon-sys` at `target/<profile>/wxdragon_sys_cmake_build/share/locale/<lang>/LC_MESSAGES/wxstd-*.mo`) and generates `wx_translations.rs` in `OUT_DIR` exposing `wx_catalog(lang)` + `wx_available_languages()`. Follows the existing `build_docs`/`readme_for_lang` codegen pattern. Embeds all 46 languages wx ships — same "embed everything present" logic as patois, avoiding any language-code remapping. Degrades with a `cargo:warning` + empty stub (so a broken build is loud, and the bundling test goes red) rather than failing.
- **`wx_translation_loader.rs`** — `WxStdCatalogLoader` (ZST) implements wxdragon's `TranslationsLoader`, handing back embedded `.mo` bytes for the `wxstd` domain. Matches both `wxstd` and `wxstd-<ver>` since wx requests the versioned domain first.
- **`translation_manager.rs`** — installs the loader via `set_loader` before `add_std_catalog`, in both `initialize()` and `set_language()` (each builds a fresh `Translations`).

## Testing

Built test-first (red → green). Tests in `wx_translation_loader.rs`:
- catalogs embedded (`.mo` magic valid, Dutch present)
- loader scoped to `wxstd` domain only
- **end-to-end**: install loader → set language `nl` → `add_std_catalog` → assert wx translates `Cancel` → `Annuleer` (headless, like wxdragon's own loader test)

`cargo test` (runs in CI) covers all of these; the bundling test is the guard against a silently-empty embed.

## Note

- Adds ~4–6 MB of `.mo` data to the binary (all 46 langs). If undesirable, the trait supports returning `Cow::Owned` decompressed bytes, so catalogs could be embedded gzipped later.


LeonarddeR:
Not sure whether the android CI failure is related.

trypsynth:
Android CI is broken on master too, stupid renovate

LeonarddeR:
In that case, I think this is ready to rock.

trypsynth:
LGTM, and it only increases the binary size by under 2MB.

PR #585, opened 2026-07-08
fix: interpret path separators per-platform in extract_title_from_path

mikolysz:
extract_title_from_path only ever receives native local paths: the desktop app and pb CLI pass the opened file's path, and Android copies content URIs to a local cache file before handing the path to the core. Treating a backslash as a separator on unix was therefore speculative generality — and actively wrong, since a backslash is a legal filename character there (a file named "dir\\" got "Untitled" instead of its name).

Gate the trailing-backslash check to Windows and split the tests into platform-independent, Windows-only, and unix-only groups. This fixes extracts_title_from_path::case_03, which asserted the Windows answer for a C:\\ path on every platform and failed on macOS.

### DIFF
--- a/crates/paperback-core/src/parser/util/path.rs
+++ b/crates/paperback-core/src/parser/util/path.rs
@@ -1,12 +1,15 @@
 use std::path::Path;
 
+/// Derives a fallback document title from the path of the file being opened.
+/// Inputs are always native local paths, so separators are interpreted
+/// per-platform: on unix, a backslash is an ordinary filename character.
 #[must_use]
 pub fn extract_title_from_path(path: &str) -> String {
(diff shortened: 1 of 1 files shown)


trypsynth:
Thanks for this! Would you be willing to make the tests run on macOS in CI in a followup PR?

mikolysz:
Sure thing!

The 4 PRs I submitted — that you’ve just merged — make the test suite green, where it previously failed to compile.

There’s some CI work to be done anyway; it seems that we try to build artifacts when contributors update master in their forks, which fails due to missing signing keys, but I’ll address that in a followup.

Thanks for the rapid response to these.

PR #618, opened 2026-07-31
fix(desktop): don't quit on arbitrary timer expiry

mikolysz:
wxdragon's `Timer::on_tick`  binds its  callback  to  the owning  frame's wxEVT_TIMER without a timer id, so  every timer's tick runs the closures of all other timers bound to the same frame. This is likely an upstream bug.

The sleep-timer closure closed the application unconditionally, which also caused it to close when any other timers fired.

This  bug was  latent on  master, as  there were  no other relevant  timers which could cause  it to trigger. However,  I have future PRS  in the pipeline that do need additional timers, hence this fix.

### DIFF
--- a/crates/paperback/src/ui/main_window.rs
+++ b/crates/paperback/src/ui/main_window.rs
@@ -549,6 +549,12 @@ impl MainWindow {
 		let dm_for_timer = Rc::clone(doc_manager);
 		let config_for_timer = Rc::clone(&config);
 		sleep_timer.on_tick(move |_| {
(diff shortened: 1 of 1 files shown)


trypsynth:
Thanks for this! However, this hack makes me a bit uncomfortable. Would it be possible for you to investigate this issue upstream in wxDragon?

mikolysz:
Sure, will do.

trypsynth:
Curious, did you get anywhere with this?

trypsynth:
a PR for wxDragon to fix this got merged and put in the newly release 0.9.19, so I think this can be closed. Please let me know if not and I'll gladly reopen it.

PR #619, opened 2026-07-31
fix(desktop): don't offer untracked tabs for Reopen Last Closed

mikolysz:
Closing any tab pushed its path onto the reopen stack, including untracked synthetic tabs like help and View Source. Reopening one of those with Ctrl+Shift+T goes through the generic open path, which knows nothing about the tab's origin: the temp file came back as a regular tracked document, titled with its raw internal filename (e.g. "chapter3.xhtml.source.txt"), and from there leaked into the recents menu, the startup-restore list (restoring a stale temp path or failing with "File not found" on the next launch), and a per-document settings entry.

Push onto the reopen stack only for tracked tabs, establishing the invariant that everything on the stack is an ordinary document that can be reopened through the generic open path.


trypsynth (approved):
LGTM, thanks!

PR #626, opened 2026-08-05
Declare support for lists in the PDF parser in supported_flags

emassey0135:
The PDF parser generates lists and list items for tagged PDFs, but right now the flag for lists is not declared in supported_flags, so navigating by list or list item will not work.


trypsynth:
Thanks for this, TBH the flag system needs improvements all around, thanks for taking this part of it on!

PR #627, opened 2026-08-06
Prepare for Audiobook Support

emassey0135:
This adds the foundation of audiobook support, both full audiobooks with no text, and books with full-text full-audio, and anything in between. The AudioTimeline struct stores audio sources (individual files), and clips, which have a source, a start and end time, and a start and end position within the text. There are also helper functions for getting the clip at a text position, getting a position from a clip, getting the percentage played in the audio (in pure audiobooks this will be vastly better than using text for progress), etc. There is also a new parser flag indicating if audio is present. To add audio to a parser, it just needs to create an audio timeline and include it in the Document. For audio-only books, I imagine the only text the parser would return is the chapter titles as headings, but this should be fine if the elapsed time is stored in bookmarks and playback position, in addition to the text position.

### DIFF
--- a/crates/paperback-core/src/audio.rs
+++ b/crates/paperback-core/src/audio.rs
@@ -1,0 +1,362 @@
+//! Maps recorded audio onto a document's text spine.
+//!
+//! An `AudioTimeline` is an overlay: it stores no text, duplicates no markers,
+//! and is rebuilt from the source document on every open, exactly like markers
+//! are. Nothing here is serialized — only `TimelinePoint` reaches disk.
+//!
+//! A point in an audio document has two coordinates, and each is authoritative
+//! for its own axis: `position` locates the text caret, `time_ms` locates the
+//! audio. Playback always resolves from `time_ms`; text navigation always uses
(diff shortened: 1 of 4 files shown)


trypsynth:
Thanks for this! I just pushed a branch I had been working on that caused some conflicts around how we handle flags though, sorry about that! Can you please fix, then I'll review?

rkingett:
Exciting! For M4B Books with many chapters, will it pull from the metadata or fallback to generic track 1, if that data isn't for some reason within the audio file?

trypsynth:
yeah, I'll make it try multiple approaches before giving up and going with the generic name.

emassey0135:
> Exciting! For M4B Books with many chapters, will it pull from the metadata or fallback to generic track 1, if that data isn't for some reason within the audio file?

When an M4B parser is written later, yes it should support chapters. This PR just adds the ability for parsers to include audio inside documents and for the future playback logic in the front-ends to reference it. The Paperback document model is all still there, so chapters will become headings if the parser extracts them from the file.

emassey0135:
It should be fixed now.

PR #628, opened 2026-08-07
Automatically reload changed documents

LeonarddeR:
Paperback parses a document once, when it is opened. If another program changes the file on disk afterwards, the open tab keeps showing the old content. The only way to see the new content is to close and reopen the document.

Readers who follow files that change — notes someone else is editing, logs, documents synced from another machine — need the open tab to pick up those changes by itself, without losing their place. Notepad++ offers this with its reload-on-refocus behavior. This change brings the same behavior to Paperback.

## What changes

- Every tab now remembers what its file looked like on disk when it was last parsed: the modification time and the size. This pair lives in a small struct called `FileFingerprint`.
- When the main window regains focus, the visible tab compares the file's current time and size against the remembered pair. The same check runs when the user switches tabs. Only the document the user is looking at is ever checked, so background tabs never cost a re-parse. The check lives in `DocumentManager::reload_tab_if_changed`.
- If the file changed, the document is parsed again and the text control is refilled in place. The reading position survives the same way the existing "render tables inline" toggle already handles it: the caret is expressed relative to the nearest document anchor before it and translated back after the new parse, with a percentage of the document as fallback. That logic was extracted from the toggle into a shared helper, `reparse_tab_in_place`, and both features use it. The navigation history is carried over to the new parse.
- After a successful reload, screen reader users hear "Document reloaded." through the live region.
- Most failures stay silent. If the file is gone or parsing fails — a half-written file, for example — the tab keeps its content and position. The failed state is remembered, so the same broken file is not re-parsed on every focus change; the next change to the file triggers a fresh attempt.
- A changed password is the exception. If the stored password no longer decrypts the file, Paperback prompts for a new one and retries the reload once, storing the new password on success. Dismissing the prompt keeps the old content in the tab without an error; the reload was not user-initiated, so nothing is lost. A wrong password shows the same error dialog as opening a document does. This lives in `DocumentManager::reprompt_password_and_reparse`.
- The behavior is on by default. A new checkbox on the General tab of the Options dialog, "Automatically reload changed documents", turns it off.

## Implementation notes

- The reload path never waits for the configuration lock. The Options dialog holds that lock while it is open, and closing the dialog re-activates the main window on the same thread; waiting there would deadlock. If the lock is busy, the check silently skips and the next focus change catches up.
- Modification time and size were chosen over the existing content-hash helper deliberately. The hash reads up to 2 MiB from disk on every check and only covers the start, end and size of a file, so it misses edits in the middle of large files. A metadata read is a single cheap system call and catches every completed write.
- Help and source-view tabs are never reloaded.

## Tests

Automated tests cover the fingerprint behavior: a missing file yields no fingerprint, an unchanged file compares equal across reads, and a rewrite with a different length compares unequal. The reload path itself needs live widgets and is verified by hand: editing an open file externally and refocusing, deleting or corrupting the file, background tabs staying untouched until selected, and turning the option off.


trypsynth:
Thanks for this! I see this is still a draft, but one question:

> • Failures stay silent. If the file is gone or parsing fails — a half-written file, a changed password — the tab keeps its content and position. The failed state is remembered, so the same broken file is not re-parsed on every focus change; the next change to the file triggers a fresh attempt.

If the password changes, could we not make it simply re-prompt? This probably wouldn't be that hard to detect I wouldn't think.
Thanks again for this!

LeonarddeR:
Implemented in 9426596. When a reload fails because the stored password no longer decrypts the file, Paperback now prompts for a new password and retries the reload once. On success the new password is stored for the document. Dismissing the prompt keeps the old content and position in the tab; nothing was lost, so no error is shown. Entering a wrong password shows the same error dialog as opening a document does.

LeonarddeR:
This is no ready for a round of review.

PR #637, opened 2026-08-10
Add "Move to start of line" option for up/down navigation

dan1982code:
## Summary

Adds a new **"Move to start of line"** checkbox to **Options → Reading**, off by default.

When enabled, pressing Up/Down places the caret at the **start** of the target line instead of preserving the current character column — matching how screen-reader virtual buffers (NVDA, etc.) navigate lines, which screen-reader users may find more predictable.

## Scope

- **Windows only.** Hooks the existing Windows-only character-column navigation handler in `document_manager.rs`. On macOS/Linux the option has no effect and the checkbox is hidden; implementing it there would be a good follow-up.
- The setting persists through the standard config system.

## Testing

- Manually tested with NVDA on Windows 11:
  - Default off: Up/Down behavior is unchanged (column-preserving).
  - Option on: Up/Down lands at the start of each line; NVDA reads each line from the beginning.
  - Setting persists across restarts.
- `cargo +nightly fmt --check`, `cargo clippy --workspace --all-targets --release`, and the release build all pass locally.
- Not covered by automated tests, since the navigation logic lives in the GUI layer.

### DIFF
--- a/crates/paperback/src/ui/dialogs/options.rs
+++ b/crates/paperback/src/ui/dialogs/options.rs
@@ -26,6 +26,7 @@ pub struct OptionsDialogResult {
 	pub start_maximized: bool,
 	pub compact_go_menu: bool,
 	pub navigation_wrap: bool,
+	pub line_start_navigation: bool,
 	pub check_for_updates_on_startup: bool,
 	pub bookmark_sounds: bool,
 	pub recent_documents_to_show: i32,
@@ -51,6 +52,7 @@ struct OptionsDialogUi {
 	start_maximized_check: CheckBox,
 	compact_go_menu_check: CheckBox,
 	navigation_wrap_check: CheckBox,
+	line_start_nav_check: CheckBox,
 	check_for_updates_check: CheckBox,
 	bookmark_sounds_check: CheckBox,
 	recent_docs_ctrl: SpinCtrl,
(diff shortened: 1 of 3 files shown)


trypsynth:
Hi @dan1982code and thanks for your contribution! This diff looks good to me! With all that said, you're a first-time contributor, so I'm going to ask: this PR was fully generated with Claude Code; did you test it? Furthermore, can you please provide me a bit of human-written justification on why I should introduce this feature? I'm not opposed to it and I see possible justification already with it being how NVDA handles it in the webview or other virtual buffers. I also understand this might be a harder ask if you're not a native English speaker, so this isn't a hard requirement for me to merge your PR. That said, while  I and the Paperback project are not opposed to agentic coding agents or AI-generated PR's, I want to make sure there's a human in the drivers seat testing it and driving the agent. Cheers, and thanks again for the pr!

dan1982code:
Hi,

Thanks! Genuine human here :) long-time computer scientist and developer, but my knowledge of Rust is zero.

I did test on Windows over here, with the latest version of NVDA. My new feature seems to be working well. But I haven't thought through the implications of ctrl+up/down or shift+up/down yet :( all I wanted to do was match the behavior of NVDA how it works in virtual buffers. Like sometimes what I do with pdf files is use Acrobat Reader to save as text, then I open the huge thing in Firefox to get it all in a virtual buffer. But with your Paperback program, I think I can finally stop doing that.

One behavioral difference I noticed was that in NVDA virtual buffers, up and down arrow always take me to the start of lines. So it's like subconscious behavior for me now, and I tried to replicate that with this feature.

I think other NVDA (maybe other screen readers, not sure about those) users might appreciate this consistency, especially if they think of your main edit area as a "virtual buffer" as I do.

I also didn't try to get any code in there for other operating systems because I have no way to test that :(

Thank you!

trypsynth:
Hi there,
Thanks a ton for this. I've tested it myself too, and it works well. I tested edgecases I wasn't sure about too, such as with word wrap enabled and ctrl+up/down, with good results. LGTM on the code-side. I'm not sure how macOS VO handles this really, so I'll leave that for the future. Thanks again, and glad Paperback hasn't yet started getting slop PR's! :)

trypsynth:
Also super glad to hear we got you off Acrobat, I'm so tired of getting emails telling me that's the only way to read damn PDF's!

dan1982code:
Can't promise no bugs but I can promise no slop from me :)

Thank you for merging... this is really motivating. Thank you for hammering on this Acrobat alternative for us. I can't be the only one who uses Acrobat just to "save as text" and then open it in Firefox. NVDA browse mode in Firefox is just so dang good. I'll try to identify other features from there to propose, too.

PR #640, opened 2026-08-11
Add temporary bookmark: / to set, \ to jump

dan1982code:
## Summary

Adds a **temporary bookmark** — a single, per-document "scratchpad" spot you can set and jump back to without cluttering the real bookmark system. My rationale is: sometimes you need to jump back and forth between a couple of places; a temporary bookmark is great for this. And it's not an important bookmark... we just want to forget about it later and not have it clutter up our real/long-lasting bookmarks. This is similar to NVDA's Place Markers addon, which has a similar temporary bookmark feature.

- **`/`** — set the temporary bookmark at the current position (overwrites any previous one)
- **`\`** — jump back to it; announces "No temporary bookmark." if none has been set. (these keystrokes, I wasn't sure which to use... I thought these made sense at least for a proof of concept)
- Per-document and persists across restarts
- Included in `.paperback` export/import, like the reading position and real bookmarks

## Design

Stored as a separate per-document config field (`temporary_bookmark`), not inside the bookmarks list. That keeps it automatically invisible to the real bookmark system:

- Not part of `b`/`Shift+b` next/previous bookmark cycling
- Not shown in the Ctrl+B / Ctrl+Alt+B bookmark lists (or any list of bookmarks)
- Never triggers the bookmark click sound

## Testing

- New core tests: set/get/overwrite/clear, serde round-trip, and old-config compatibility (missing field loads as none). Full `paperback-core` suite: 509 passing.
- `cargo +nightly fmt --check`, `cargo clippy --workspace --all-targets --release`, and the release build all pass.
- Manually tested with NVDA on Windows (set, jump, no-temp-bookmark case, persistence across restart, and confirmed it stays out of bookmark lists/navigation).
- Don't know if it works on all platforms (the key handler is cross-platform but I can't test).

### DIFF
--- a/crates/paperback-core/src/config.rs
+++ b/crates/paperback-core/src/config.rs
@@ -213,6 +213,8 @@ pub struct DocumentConfig {
 	pub navigation_history_index: usize,
 	#[serde(default)]
 	pub bookmarks: Vec<StoredBookmark>,
+	#[serde(default, skip_serializing_if = "Option::is_none")]
+	pub temporary_bookmark: Option<i64>,
 	#[serde(default, skip_serializing_if = "String::is_empty")]
 	pub format: String,
 	#[serde(default, skip_serializing_if = "String::is_empty")]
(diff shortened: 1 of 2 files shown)


trypsynth:
Can you please resolve the merge conflicts? LGTM other than that, thanks! Wouldn't have thought of this.

dan1982code:
Oops -- sorry :( just tried to fix it. Thank you!

PR #648, opened 2026-08-15
Stop creating a macOS home-directory lock file

johanntan:
## Summary

- disable wxWidgets single-instance lock handling on macOS
- retain the existing single-instance and IPC behavior on Windows and Linux
- compile platform-specific IPC declarations only where they are used

## Why

Paperback passed no lock directory to wxDragon's `SingleInstanceChecker`. On Unix platforms, wxDragon defaults that lock to the user's home directory, causing the mac client to create `~/paperback_running`.

macOS already delivers reopen and document-open events to the running application through native application events. Paperback also has no macOS command-forwarding transport, so retaining the cross-platform lock checker there created an unwanted home-directory file without providing functional IPC forwarding.

## User impact

The mac client no longer creates `~/paperback_running`. Windows and Linux single-instance behavior is unchanged.

## Validation

- `cargo check -p paperback`
- `cargo test -p paperback ipc::tests` (10 passed)
- project pre-commit hooks, including nightly `cargo fmt`

### DIFF
--- a/crates/paperback/src/ipc.rs
+++ b/crates/paperback/src/ipc.rs
@@ -3,11 +3,14 @@ use std::{
 	path::{Path, PathBuf},
 };
 
+#[cfg(any(target_os = "linux", target_os = "windows", test))]
 pub const IPC_COMMAND_ACTIVATE: &str = "ACTIVATE";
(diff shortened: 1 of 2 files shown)


johanntan:
I just wanna add Codex was used to make the changes and this PR.

PR #659, opened 2026-08-23
Updated Finnish UI texts

jkinnunen:


trypsynth:
@jkinnunen do you actually listen to these codex suggestions? In other words, should I leave this PR open until you resolve the conversation, or is it just noise?

jkinnunen:
I only read them occasionally, but I still think they’re useful for translators.

PR #667, opened 2026-08-26
Use `gettext` plural forms for counted strings

tareh7z:
Several status bar and dialog messages that include a count (document totals, sleep timer minutes, word and character counts, estimated reading time) hard-coded a single plural word or built the plural form by linking the number with plain text, which only works for English and breaks languages like Bosnian or Serbian that need three plural forms. This switches those messages to patois's existing `ngettext-style nt()` function.


trypsynth:
Thanks, super useful now that I made patois support them :)

PR #710, opened 2026-08-29
Update Simplified Chinese translations in

piano777wzy:
Updated translation file for Simplified Chinese, including changes to message strings and plural forms.


piano777wzy:
Hello, I don't know if this pull request is correct, I changed zhcn on the webpage.The po file is then followed step by step

PR #726, opened 2026-08-30
Find: in-window search strip (no verbose screen-reader focus announcement)

dan1982code:
# Find: in-window search strip (no verbose screen-reader focus announcement)

OK, this proof of concept might be too heavy a change to fix what I am hoping to fix, but...

Fixes an accessibility annoyance with Find, stemming from it being a separate dialog window inside the main window.

## The problem

The small annoyance here that I'm trying to fix is that Closing Find made NVDA announce the whole focus chain. Pressing Enter (or Cancel) hid the dialog, focus returned to the book text, and NVDA read "Paperback, book title, tab control, edit read only multi line" every time. That's standard NVDA behavior when focus moves across a window boundary — the dialog is a *separate window*, so the focus return is reported with the full ancestor chain. It's most disruptive for Find, where you just want to hear the found text.

(Incidentally, I don't know how it does it, but when you do NVDA find in a web browser, ctrl+NVDA+f, there's no extra verbeage spoken after you search for something. I don't know how they avoid re-reading the application title bar.)

## What changed

**Find is now an in-window strip, not a dialog.** It's a bordered panel with a "Find" heading that lives inside the active tab's page and floats over the book text, centered — so it still *reads* as a dialog (Ctrl+F jumps focus to the find combo). Because it shares the book's window, focus moving between the find controls and the text is a within-window change: NVDA no longer announces the "Paperback, tab control" preamble on the way back. It's rebuilt under the current page when you switch tabs. (Visually, I have no idea how this looks :D )

- **Found text first.** A successful search announces the line containing the match *before* focus returns, so you hear the result, then (at most) a brief "edit, multiline".
- **Escape closes it**, handled on the combo, the panel, and a notebook-level fallback.
- **Closing restores focus to the book text.** A hidden panel's combo keeps keyboard focus — a wxWidgets/Windows quirk that made Escape keep landing on the invisible strip — so `hide()` explicitly returns focus to the text control, which also repaints the area the strip covered.
- **Delete inside the strip edits its text** rather than triggering the tab-strip's close-document gesture.

**Single-key shortcuts are no longer OS accelerators.** `format_menu_label` now shows single-key shortcuts inline (`Next Table (T)`) instead of as a tab-aligned `\tT`. The `\t` form is what registered them as Windows accelerators; dropping it means they can't steal keystrokes from any in-window text field. They're now handled by the existing text-control key dispatch, so they still work while reading the book but never while typing. Modifier shortcuts (Ctrl+F, Ctrl+W, Ctrl+Space, …) keep the `\t` format and stay real accelerators. This fixes the collision class for any future in-window editable control, not just Find. (This also has the side effect that stuff like t, l, etc. doesn't work when focused on the tab strip anymore. That may be more expected behavior compared to still firing on the active book, but I'm not sure.)

## Notes / tradeoffs

- **Menu display changes**: single-key shortcuts now read inline as "Next Table (T)" rather than relying on tab-aligned layout.
- **Single-key navigation now requires the book text to be focused** (previously the accelerator fired from anywhere in the window, e.g. the tab strip).
- **Only Find gets the in-window treatment here.** The other dialogs (Go to Line, Sleep Timer, …) are still modal windows and keep the standard NVDA focus-return announcement; that's a per-dialog decision for a future pass.

## Testing

Tested with NVDA on Windows 11: Ctrl+F pops up the strip with the combo focused (no chain); letters type normally; Enter announces the found line and returns focus to the book without the "Paperback, tab control" preamble; Escape/Cancel close it cleanly; single-key and Ctrl shortcuts still work from the book text; Alt+Tab focus restore is unaffected.

### DIFF
--- a/crates/paperback/src/ui/find.rs
+++ b/crates/paperback/src/ui/find.rs
@@ -51,7 +51,16 @@ pub fn find_text_with_wrap(haystack: &str, needle: &str, start: i64, options: Fi
 
 #[derive(Clone)]
 pub struct FindDialogState {
-	pub dialog: Dialog,
+	/// The in-window find strip. A child of the active tab's page rather than a separate dialog
+	/// window, so focus moving between the find controls and the book text stays within the same
+	/// top-level window and NVDA doesn't announce the whole "Paperback, tab control" chain on the
+	/// way back. Shown/hidden (with a border and a heading) so it still reads as a dialog.
+	pub panel: Panel,
+	/// The tab page the strip was created under, so it can be rebuilt if the active tab changes.
+	host_page: Panel,
+	/// Used to return focus to the book text when the strip hides: a hidden panel's combo keeps
+	/// keyboard focus, so without this ESC keeps landing on the (invisible) find strip.
+	doc_manager: Rc<Mutex<DocumentManager>>,
 	find_combo: ComboBox,
 	match_case: CheckBox,
 	whole_word: CheckBox,
@@ -62,13 +71,15 @@ pub struct FindDialogState {
 impl FindDialogState {
 	pub fn new(
 		frame: &Frame,
+		host_page: Panel,
 		config: &Rc<Mutex<ConfigManager>>,
 		doc_manager: &Rc<Mutex<DocumentManager>>,
 		find_dialog: &Rc<Mutex<Option<Self>>>,
 		live_region_label: StaticText,
+		initial_text: &str,
 	) -> Self {
-		// TRANSLATORS: Title of the Find dialog
-		let dialog = Dialog::builder(frame, &t("Find")).build();
+		// TRANSLATORS: Heading shown at the top of the Find panel
+		let panel = Panel::builder(&host_page).with_style(PanelStyle::TabTraversal | PanelStyle::BorderRaised).build();
 		let FindDialogWidgets {
 			find_combo,
 			match_case,
@@ -77,10 +88,10 @@ impl FindDialogState {
 			find_prev_btn,
 			find_next_btn,
 			cancel_btn,
-		} = build_find_dialog_ui(dialog);
+		} = build_find_dialog_ui(panel);
 		bind_find_dialog_actions(FindDialogActionParams {
 			frame: *frame,
-			dialog,
+			panel,
 			find_combo,
 			find_prev_btn,
 			find_next_btn,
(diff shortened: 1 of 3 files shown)


trypsynth:
Thanks for this! I want to put out 0.9.1 today and this is admittedly a bit of a big change for that, but I'll look at this soon and also see if I can dig into NVDA's code to see what they're doing for the find dialog to make it not do that. Thanks as always for the contribution!

dan1982code:
Thank you! If you're like, "why are you trying to super-optimize NVDA feedback?", it's because Paperback is so darn streamlined already that I am inspired to help push this to screaming top speed.

Thank you for an amazing app.

trypsynth:
I went and read NVDA's source, and it turns out NVDA does not avoid the focus chain announcement at all. It interrupts it.

`source/cursorManager.py`, `FindDialog.onOk`:

```python
def onOk(self, evt):
    text = self.findTextField.GetValue()
    caseSensitive = self.caseSensitiveCheckBox.GetValue()
    core.callLater(100, self.activeCursorManager.doFindText, text, caseSensitive=caseSensitive, reverse=self.reverse)
    self.Destroy()
```

The dialog is destroyed, focus returns, NVDA starts speaking the whole chain, and 100ms later `doFindText` runs, calls `speech.cancelSpeech()`, and speaks the found line. You never hear the preamble because it gets cut off mid word. Same architecture we already have, just with an interrupt on the end.

The good news is that NVDA exposes that same lever to external apps. From `NVDAObjects/UIA/__init__.py`:

```python
if notificationProcessing in (
    UIAHandler.NotificationProcessing_ImportantMostRecent,
    UIAHandler.NotificationProcessing_MostRecent,
):
    if speech.sayAll.SayAllHandler.isRunning():
        speechPriority = speech.priorities.Spri.NOW
    else:
        speech.cancelSpeech()
ui.message(displayString, speechPriority=speechPriority)
```

So a UIA notification event with `NotificationProcessing_ImportantMostRecent` is `cancelSpeech()` plus speak, driven from outside NVDA.

Rather than trust a source read I built a spike: a Win32 window that raises notification events, plus a second process running a bare UIA client subscribed to notification events, so I could tell "NVDA dropped it" apart from "it never entered the UIA tree". Results on NVDA 2026.1, Windows 11 24H2:

| processing | behaviour |
| --- | --- |
| ImportantMostRecent (1) | interrupts speech in progress, then speaks |
| MostRecent (3) | interrupts |
| ImportantAll (0) | queues |
| All (2) | queues |

The control case is what makes this trustworthy. With `ImportantAll` the queued long announcement ran all the way to the end before the second one was spoken, and with the two interrupting values it got cut off partway through, so the cutoffs are real and not a coincidence of timing.

One trap that cost me a few rounds: `UiaRaiseNotificationEvent` called on the provider from `UiaHostProviderFromHwnd` returns S_OK and reaches no UIA client at all. No error, no event, nothing. The window has to expose a real server side `IRawElementProviderSimple` from `WM_GETOBJECT` / `UiaRootObjectId`, returning the host provider from `HostRawElementProvider`. Once that was in place all seven test events routed with the correct processing values and NVDA behaved exactly as the table says.

Worth noting for context: our current live region path (`EVENT_OBJECT_LIVEREGIONCHANGED` with a polite `LiveSetting`) has no interrupt available at all. That is a property of MSAA live regions, not of the announcement, and it is why announcing the found line before moving focus was the only ordering that worked for you.

So the idea is to keep Find as a real `wxDialog` and, on close, raise an interrupting notification carrying the found line about 100ms later, mirroring NVDA's own timing. That gets the result you were after, keeps `role=dialog`, tab trapping, and escape and default button handling, and it fixes the identical annoyance for Go to Line, Sleep Timer and the rest instead of just Find. During say all NVDA uses `Spri.NOW` rather than cancelling, so continuous reading does not get chopped up either.

The cost is a dedicated announcer HWND implementing that provider, roughly 60 lines of COM, probably living in the live-region crate, plus a one shot timer. The document control stays on IAccessible so book reading is untouched. Caveat: I have only verified NVDA so far, JAWS and Narrator notification semantics are untested.

Separately, the accelerator change here is more interesting than the workaround framing suggests. The text control key handler in `document_manager.rs` already resolves every `ActionId` and falls through to posting the menu command, so single key shortcuts were being handled twice with the OS accelerator winning. Dropping the `\t` is genuine de-duplication. I would want it as its own PR though, since the inline `(T)` label changes every translated menu string and losing single key nav from the tab strip is a visible behaviour change worth discussing on its own.

I am going to hold this one while I try the notification approach, but thank you for it. Your diagnosis of the root cause was spot on, and chasing it down taught me a lot about how NVDA actually handles this.

dan1982code:
Thank you for this! I could give it a try over here, new PR, and see if I can get it to work?

trypsynth:
I just published version 0.3.0 of my live-region crate, about to update Paperback to use it. Now priorities are properly supported, and you can cut off the currently speaking speech with another message. Once that gets pushed, please feel free to open a fresh PR to try this :) thanks for getting me going down this rabithole!

dan1982code:
On it now :) thanks!

dan1982code:
Thanks for merging!

We should let people test before I keep going with some other dialogues like go to line?

Also I just noticed that f3/shift+f3 wraps even when the wrap navigation checkbox in settings is unchecked. Is that intentional?

PR #727, opened 2026-08-30
update the latest Simplified Chinese language.

piano777wzy:
update the latest Simplified Chinese language.
I noticed that there is a lot of Android-related content, but the latest Android version still cannot use other languages. It remains in English by default. Also, the Help option in More Settings cannot be opened, and it says the file cannot be found. Android also does not have a language switching option in Settings like Windows does.


trypsynth:
Oh shoot, it looks like Claude is still machine translating this language. To fix this, either add the language code to the txt file in the po folder, or I can do it. Sorry about that!

piano777wzy:
> Oh shoot, it looks like Claude is still machine translating this language. To fix this, either add the language code to the txt file in the po folder, or I can do it. Sorry about that!


I don't know what's going on, but this is a complete translation.

PR #730, opened 2026-08-30
Repair the Polish translation and bring its help up to 0.9.0

michaldziwisz:
The auto-translate runs had filled the gaps in this locale with machine output that `msgfmt` accepts but a user does not. Since `pl` is on the human-maintained list the bot will not touch it again, so this is a one-time cleanup of the damage already committed rather than something that will keep recurring.

The catalog goes from 335 translated / 449 fuzzy / 52 untranslated to **836 translated, nothing fuzzy or empty**. Every entry was reviewed by hand against the current pot. Worth noting that fuzzy entries are not harmless here: `patois-build` compiles with `msgfmt --use-fuzzy`, so all of these were reaching users.

### What the machine had broken

- **13 menu accelerators lost**, because the ampersand was translated as the Polish word *i* (and): "Find &Next" became "Znajdź i następny" (*Find and next*), "Go &Back" became "Przejdź do przodu i do tyłu" (*Go forward and back*), "Reveal &File in Folder" became "Pokaż i zapisz w folderze" (*Show and save in folder*).
- **6 entries lost the tab** separating a label from its shortcut, replaced with spaces, so a screen reader announces `Zaznacz wszystko    Ctrl+A` as the item's name rather than as a shortcut.
- **Plural forms broken**: ngettext entries dropped `%d` entirely (`%d document.` became just "dokument") or used the wrong placeholder syntax, and all three Polish plural forms were identical, which defeats the point of having three.
- **Semantic errors**: Undo and Redo shared one translation, `Cu&t` was left in English, `Font: {}` became "Źródło: {}" (*Source: {}*), `Go` became "Start".

### Also in this PR

Terminology unified across the catalog (Web View was rendered two different ways, `parser` two ways, and the word "word" two ways), and two wording fixes: the signature failure message said the update might have been *modified*, losing the security sense of "tampered with", and the image-only PDF notice turned "may need" into a categorical "must".

**The help was still the 0.8.5 document** — 30 headings against the English 64 — and it would never have corrected itself, since listing `pl` in `human-maintained-locales.txt` opts it out for `doc/readme-<lang>.md` too. The most serious omission was the **Screen Reader Compatibility** section, including the JAWS and Braille Displays subsection with the `paperback.jcf` workaround: in an app written for screen reader users that is the section that matters most, and a reader using the Polish help had no way to learn about it. Also added: the 0.9.0 changelog, the new system requirements, four newly supported file types, the macOS shortcut equivalents, the container navigation and temporary bookmark keys, Dutch in the language list, and one donor name. The 0.8.5-and-older changelog was already complete and is untouched.

Key names in the help are now Polish while modifiers are left alone, matching what the German translation does (`Leertaste`, `Alt+Pfeil nach links`, `Entf auf dem Ziffernblock`): `Ctrl+Space` → `Ctrl+Spacja`, `Alt+Left` → `Alt+Strzałka w lewo`. Ctrl, Alt and Shift are printed that way on Polish keyboards, so translating those would help nobody.

### Checks

`msgfmt --check`: 836 translated, 0 fuzzy, 0 untranslated. Placeholders, accelerator counts, shortcut tabs and the three plural forms verified per entry. Help structure verified against `doc/readme.md`: same headings at the same levels, same bullet count per section, identical shortcuts and file extensions. `cargo test -p paperback-core`: 628 pass; the 2 failures are identical on unmodified `master` (table display-length tests), so they are not from this change.

Reviewed with several independent reviewers and kept only findings that held up against the file and the code — for instance "loaded document window" is the loaded buffer rather than a UI window, so the Polish now says *fragment* rather than *okno*.

Happy to split, reword or drop any part of this if you'd prefer it shaped differently.

### DIFF
--- a/doc/readme-pl.md
+++ b/doc/readme-pl.md
@@ -1,4 +1,4 @@
-# Paperback - wersja 0.8.5
+# Paperback - wersja 0.9.0
 
 ## Wprowadzenie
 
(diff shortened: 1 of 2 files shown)


trypsynth:
Thanks, good stuff!

PR #742, opened 2026-08-31
Polish: translate the new strings, and split the plural forms the marker allows

michaldziwisz:
The pot grew in #737, so the Polish catalogue went from complete to 836 of 847. This brings it back to 847.

**The four reworded strings keep their existing translations** — `&Locate…` to `&Locate...`, `No next section` to `No next section.`, and `1 of 1 document selected.` became `1 document selected.`, which reads better in Polish too.

**The plural part is what I would like you to look at.** Polish picks between three forms the same way `nt()` does, so *2 sekundy* and *5 sekund* are genuinely different words. Before this commit the catalogue had a single entry for both, translated with the "many" form, so every count ending in 2, 3 or 4 was announced wrong: *2 sekund* where a reader expects *2 sekundy*. Your `PLURAL_MANY_MARKER` is exactly what fixes it, by giving "few" and "many" separate keys:

| msgid | Polish | counts |
|---|---|---|
| `{} second` | `{} sekunda` | 1, 21, 31, … |
| `{} seconds` | `{} sekundy` | 2-4, 22-24, … |
| `{} seconds` + U+2063 | `{} sekund` | 0, 5-20, 25-30, … |

Same for `Custom: {} minute(s)` in the iOS sleep timer. As far as I can tell Polish is the first catalogue to actually use the split — Bosnian, French, Spanish and Russian all give the two keys identical text, so the marker currently buys them nothing.

The marker is not repeated in the translations, per your comment on `PLURAL_MANY_MARKER`: translators write the text only.

**One thing I did not change, because it is a code change rather than a catalogue one.** In `WordCountDialog.kt` the word count passes the same msgid as both `few` and `many`:

```kotlin
nt(t("This document contains {} word."), t("This document contains {} words."), t("This document contains {} words."), stats.wordCount)
```

so Polish cannot distinguish *2 słowa* from *5 słów* there — both come out as whichever form the single entry holds. Appending the marker to the third argument would let it, exactly as the seconds case already does. Same for the bare `nt(t("word"), t("words"), t("words"), …)` just below it. Happy to send that as a separate PR if you would like it.

### Verification

- `msgfmt --check --statistics`: 847 translated, no format errors
- placeholder and marker audit over all 847 pairs: no mismatches, no marker leaking into a translation
- ran the `nt()` rule over counts 1-111 against the new catalogue: all twelve sampled forms read correctly in Polish, where the previous catalogue got four of them wrong (2, 3, 4, 22)
- diffed against master entry by entry: no existing translation lost, and the only two changed are the deliberate "few" corrections above

### DIFF
--- a/po/pl.po
+++ b/po/pl.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: paperback 0.8.5\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-08-29 11:48-0600\n"
+"POT-Creation-Date: 2026-08-30 23:46+0200\n"
 "PO-Revision-Date: 2026-06-08 13:45+0200\n"
 "Last-Translator: Michał Dziwisz <michal@dziwisz.net>\n"
 "Language-Team: Polish\n"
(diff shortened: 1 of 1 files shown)


tareh7z:
Hi @michaldziwisz!

Good catch, I've fixed the `WordCountDialog.kt` many-form in #746.

Thanks.

PR #753, opened 2026-08-31
Polish: translate the word-count many-forms from #746

michaldziwisz:
Thanks for fixing `WordCountDialog.kt` in #746 — this fills in the Polish side of it.

Two new keys had no Polish text, so the word-count dialog fell back to English for counts like 5 or 25. With the marked key now carrying the "many" text, the unmarked `This document contains {} words.` is the "few" form (2, 3, 4, 22, ...), so it needed "słowa" instead of "słów".

`msgfmt --check --statistics`: 849 translated, 0 untranslated, 0 fuzzy. No other entry changed; the only other diff is the POT date and a TRANSLATORS comment that #748 dropped from the pot itself.

While verifying this I found a separate bug that a catalogue change cannot fix, so I'm reporting it rather than sneaking it in here: `nt()` in `Translations.kt` picks the "one" form for any count ending in 1 except 11, which is right for Bosnian, Serbian and Russian but wrong for Polish — Polish uses the singular for 1 only, so the dialog currently says "21 słowo" and "121 słowo" where it should say "21 słów". Measured by compiling the real `nt()` body and comparing its choice against `gettext` on `pl.po` for 0–10000: 899 counts disagree, while Serbian and Russian disagree on none. Happy to open a code PR if you'd like a fix.

### DIFF
--- a/po/pl.po
+++ b/po/pl.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: paperback 0.8.5\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-08-30 23:46+0200\n"
+"POT-Creation-Date: 2026-08-31 15:15+0000\n"
 "PO-Revision-Date: 2026-06-08 13:45+0200\n"
 "Last-Translator: Michał Dziwisz <michal@dziwisz.net>\n"
 "Language-Team: Polish\n"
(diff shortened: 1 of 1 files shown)


trypsynth:
Sure, a fix for that would be nice if you're willing. Thanks for this! Also, the 0.9.2 changelog jsut landed, although I'm not shipping it for at least a few hours, if you want to update that.

PR #754, opened 2026-08-31
Take nt()'s plural rule from the catalogue instead of hardcoding one

michaldziwisz:
This is the code fix for the `nt()` bug I mentioned in #753, kept separate from the catalogue change there so you can take either on its own.

`nt()` hardcodes the Bosnian/Serbian rule — "one" for any count ending in 1 except 11 — which is right for the language the marker mechanism was added for, but wrong for other three-form languages. Polish gives the singular to 1 alone, so the word-count dialog announces "21 słowo" and the seek label reads "21 sekunda" where the language wants "21 słów" and "21 sekund".

The rule each language wants already ships with its translations: `gen_android_strings` writes the po header into the asset under the empty msgid, the same entry desktop reads through patois. So `Translations.load` parses its `Plural-Forms` line and `nt()` uses it, with the hardcoded rule kept as a fallback when there's no header or it doesn't parse — behaviour is unchanged wherever there's nothing better to go on.

It turned out not to be only Polish. Two-form and one-form languages were reaching a slot their catalogue leaves empty: with a Finnish or Chinese catalogue loaded, a count of 0 or 5 landed on the marked "many" key, which those languages don't translate, and the label fell back to English.

Measured for every shipped catalogue by comparing what `nt()` selects against gettext's own choice for 0–10000, using each language's real header:

| catalogue | visible defects before | after |
|---|---|---|
| pl | 899 | 0 |
| de | 899 | 0 |
| fi | 7300 | 0 |
| zh_CN | 6401 | 0 |
| bs, es, fr, nl, pt_br, ru, sr, vi | 0 | 0 |
| cs, ja (no `Plural-Forms` header) | unchanged from the hardcoded rule | |

"Visible defect" counts a count whose displayed string differs, not merely a different slot: for a two-form language both keys usually hold the same text, so those slot changes are invisible and aren't counted as fixes.

Also checked against rules Paperback doesn't ship yet, since catalogues keep being added — Czech's parenthesized form, Lithuanian, Latvian, Welsh, Slovenian, Irish (5 forms) and Arabic (6) all agree with gettext, and a language with more forms than `nt()` has slots clamps to the last rather than throwing.

`:app:testDebugUnitTest` is green: 46 tests, 0 failures, including the 11 new ones. I checked they can actually fail — reverting `nt()` to ignore the header fails exactly 4 of them, while the three fallback tests stay green, as they should.

The parser is deliberately narrow (`?:`, `||`, `&&`, comparisons, `%`, integer literals — the whole of what gettext plural rules use) and returns null rather than guessing, so a malformed header can't silently select form 0 for every count.

### DIFF
--- a/android/app/src/main/kotlin/dev/paperback/android/Translations.kt
+++ b/android/app/src/main/kotlin/dev/paperback/android/Translations.kt
@@ -172,12 +172,14 @@ internal class PluralRule private constructor(
 				?.lineSequence()
 				?.firstOrNull { it.startsWith("Plural-Forms:") }
 				?: return null
-			val nplurals = line.substringAfter("nplurals=", "")
+			val nplurals = line
+				.substringAfter("nplurals=", "")
 				.takeWhile { it.isDigit() }
 				.toIntOrNull()
 				?.takeIf { it > 0 }
 				?: return null
-			val expression = line.substringAfter("plural=", "")
+			val expression = line
+				.substringAfter("plural=", "")
 				.substringBefore(';')
 				.trim()
 				.takeIf { it.isNotEmpty() }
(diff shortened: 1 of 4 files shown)


trypsynth:
Thanks, LGTM! At some point I might want to merge some of this into patois, but that's its own issue.

PR #756, opened 2026-08-31
Polish: translate the 0.9.2 strings and changelog

michaldziwisz:
Polish is up to date for 0.9.2, in time for tonight's release.

The five new strings are the audio-only document ones: the three Document Info labels, the file-count sentence, and the Word Count duration line. `msgfmt --check`: 854 translated, 0 untranslated, 0 fuzzy.

Also in here: the 0.9.2 changelog section, plus two bullets dropped to match "Some readme tweaks" from earlier today. The Polish file still carried translations of the bookmark-sounds and dev-builds entries you removed from the English changelog, which is why its 0.8.5 and 0.8.0 sections had one bullet more than yours. Our structure check now reports 66 headings each, 110 backticked items each, and no bullet-count mismatch anywhere.

The title still reads 0.9.1, matching `doc/readme.md` — happy to bump both if you'd rather they say 0.9.2 before the release.

### DIFF
--- a/doc/readme-pl.md
+++ b/doc/readme-pl.md
@@ -202,6 +202,21 @@ Uwaga: publiczne sponsorowanie w GitHub traktuję jako podstawę do automatyczne
 
 ## Lista zmian
 
+### Wersja 0.9.2
+* Audiobooki nie każą już czytnikowi ekranu odczytywać ciągu spacji, kiedy przejdziesz do pola tekstowego.
+* Audiobooki podają teraz nazwę pliku, kiedy przechodzisz między nimi po sekcjach.
+* Audiobooki pokazują teraz swoją prawdziwą długość, zamiast twierdzić, że każdy plik trwa 24 godziny.
+* Zamknięcie widoku WWW klawiszem Escape nie wyświetla już komunikatu diagnostycznego, kiedy wcześniej otworzyłeś w nim odsyłacz.
+* Kopiowanie po zaznaczeniu wszystkiego daje teraz cały dokument, a nie tylko tę jego część, która jest właśnie wczytana.
+* Wyszukiwanie przenosi teraz od razu do znalezionego wiersza, bez wysłuchiwania, jak czytnik ekranu ponownie odczytuje całe okno przy powrocie do książki.
+* Naprawiono otwieranie plików EPUB z pozostawionym blokiem ZIP64, które kończyło się komunikatem „Invalid local file header".
+* Naprawiono wracanie długich dokumentów na początek, kiedy czytnik ekranu czytał je ciągiem.
+* Odsyłacze w widoku WWW prowadzą teraz do wskazanej sekcji, zamiast kończyć się komunikatem „Nie znaleziono pliku".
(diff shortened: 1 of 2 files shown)


michaldziwisz:
Heads up before tonight's release: the two `Tests` failures here aren't from this PR, and they aren't Polish-specific.

`duration_format`'s tests assert the English text (`"2 hours"`, `"1 hour, 1 minute, 5 seconds"`), but `format_duration_seconds` goes through `nt()`, so patois translates by the *machine's* locale. Both runners came back French: `left: "2 heures"`, and `left: "1 heure, 1 minute, 5 seconds"` — note that one is half-translated, since fr has no `%d seconds` entry. Same failure on master's own CI run for `Some readme tweaks`, before this branch existed.

Checked it can't be us: this PR touches only `po/pl.po` and `doc/readme-pl.md`, and `git log -S` shows no commit here touching `%d hour`, `%d minute` or `%d second` — those Polish entries are unchanged from master. A Polish machine would fail the same tests the same way ("2 godziny"), and did so before today too.

Everything else is green: Android checks, Clippy, Format check, and all four builds.

If it's useful, the fix is to make the tests independent of the ambient locale rather than to change any catalogue — e.g. asserting on a segment count/shape, or forcing the locale to English for that test.

trypsynth:
Thanks, this looks good. `msgfmt --check` gives 854 translated, 0 fuzzy, 0 untranslated against 854 in the pot, and the 0.9.2 section carries all 13 bullets in the same order as the English list.

On the title: I have bumped `doc/readme.md` to 0.9.2 on master (fd462e12), so please bump `doc/readme-pl.md` to match before this merges.

On the two `Tests` failures, you are right that they are not from this PR and not Polish specific, but the cause is not the machine's locale. `dependency_owned_strings_translate_through_the_app_catalog`, in `crates/paperback/src/translation_manager.rs`, sets the patois locale to `fr` and restores `en` at the end of the test. patois keeps the locale in a process global registry and cargo runs tests as threads in one process, so anything running in that window sees French. Measured against the release binary: `duration_format` alone fails 0 times in 30, and run alongside that test it fails 3 times in 30, with `left: "2 heures"` and `left: "0 seconde"`.

The output you quoted is what settles it. `fr.po` on your branch head does have `msgstr[1] "%d secondes"`, so an ambient French locale would have produced fully French text. `"1 heure, 1 minute, 5 seconds"` is only possible if the locale flipped between the `nt()` calls inside a single `format_duration_seconds` call, which is the race caught mid function. It surfaced today because those tests arrived a few hours ago in fee1558b.

Pinning `en` inside `duration_format` would still race against the same global, so I am putting a shared lock around the locale touching tests instead, with a panic safe restore. Test only, so it does not hold up tonight's release.

trypsynth:
Merging as is: the two Tests failures are the locale race, now fixed on master in 3b6b9274, and this branch predates that fix. I am also bumping the readme titles for every language in one commit right after, so no need to touch the Polish title yourself. Thanks for the quick turnaround.

PR #762, opened 2026-09-01
Add Linux build (AppImage + tar.gz) and desktop integration

hexxbyte:
Adds a native Linux build, distributed as a portable tar.gz and a self-contained AppImage.

## What's here

* CI: new `build-linux` (x64) and `build-linux-arm64` jobs in `build.yml`, wired into `release` alongside the existing Windows/macOS/Android jobs. Installs GTK3, WebKitGTK, and the GStreamer dev packages `wxdragon`'s Linux `media-ctrl` feature needs to link.
* Packaging (`xtask`): builds `paperback.tar.gz` (just `paperback`, `pb`, and `libpdfium.so`, flat, no wrapper directory) and `paperback.AppImage` via `appimagetool`. The AppImage build is best-effort — a missing `appimagetool` skips it with a warning rather than failing the release.
* pdfium: `build/pdfium.rs` now fetches/builds `libpdfium.so` for Linux the same way it already does for Windows/macOS.
* First-run setup dialog (`linux_integration.rs` + `ui/dialogs/linux_setup.rs`): since an AppImage has no installer to hang a Tasks page off of, Paperback shows an equivalent checkbox list on first launch — the same format associations the Windows installer offers (via `paperback_formats::ALL`), applied through `xdg-mime` and a `paperback.desktop` file written to `~/.local/share/applications`, plus an "Add `pb` to PATH" option that symlinks it into `~/.local/bin`. No-ops entirely for non-AppImage installs (`.deb`/`.rpm` packagers are expected to ship their own desktop file and associations). Cancel/Escape quits rather than opening a window the user never asked for, and leaves the "already set up" flag unset so it's offered again next launch.
* Config/logs: stored under `$XDG_CONFIG_HOME/Paperback` on Linux, matching the existing per-platform convention.
* Website: `web/index.md` and `web/js/downloads.js` updated to list Linux alongside Windows/macOS, with x64/ARM64 download links for both the AppImage and the tar.gz.


hexxbyte:
I don't think the CI failure is related to this PR, but please let me know if it is, and I can fix it.

PR #764, opened 2026-09-01
Open M4B audiobooks as chaptered audio-only books

johanntan:
Paperback already has an audio timeline and native playback for direct local sources, but standalone .m4b files stop at format routing. This adds an M4B parser backed by mp4ameta and reads metadata without pulling embedded cover bytes into memory.

Both MP4 chapter lists and chapter tracks become normalized clips with named section and table-of-contents anchors. A blank line per chapter gives the audio-only document a caret spine, matching plain-audio ZIP books; empty chapter titles get a translated fallback, and chapterless files open as one full-duration section.

The shared format table now carries .m4b with audio, section and TOC support, so desktop associations, file dialogs, the CLI and mobile extension filters all receive it. The English format lists and translation template are updated alongside it; .m4a remains unclaimed.

Verified with cargo test -p paperback-core --no-default-features, cargo test -p paperback-formats, cargo check --workspace, the focused seven-test M4B suite, Clippy under the repository lint policy, and an end-to-end pb smoke test against a real MPEG-4 audiobook.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1559,6 +1559,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "mp4ameta"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "453e991b2efe96c288cbc2d03a19e353504294559ecb6c03067fff5bd824616d"
+
+[[package]]
 name = "native-tls"
 version = "0.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1777,6 +1783,7 @@ dependencies = [
  "icu_properties",
  "libchm",
  "lofty",
+ "mp4ameta",
  "office-crypto",
  "paperback-formats",
  "patois",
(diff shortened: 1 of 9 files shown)


trypsynth:
LGTM! One quick question, is there a reason you used mp4ameta? I'm not at all opposed to dragging in new libraries, but lofty, which we already depend on, I think?  Can do this? Feel free to correct me if I'm wrong, like I said I'm not at all opposed to introducing new dependdencies, just wonder if this one is strictly necessary. Thanks again!

johanntan:
Hi,
I, as you probably know already completely use AI to make these, and here is what it had to say on why it used it.


Good question! I checked Lofty before adding mp4ameta. Lofty can read the M4B’s duration and ordinary MP4 ilst metadata, but it doesn’t currently expose MP4 chapters. Its chapter-frame support is ID3v2-specific, whereas M4B chapters may be stored either in a chpl atom or a QuickTime chapter track. mp4ameta supports both representations, which is what the parser uses to construct the TOC and audio timeline. We could use Lofty for the other fields, but we would still need mp4ameta for chapters and would end up parsing the file twice.

trypsynth:
Oooh okay, that makes sense. Thanks, lGTM.

PR #769, opened 2026-09-01
Go to percent: land on the entry, cut the focus chain, reject bad percentages, and accept +n/-n

dan1982code:
# Go to percent: land on the entry, cut the focus chain, reject bad percentages, and accept +n/-n

Improvements to the Go to percent dialog (Ctrl+Shift+G, or Go → "Go to percent"), parallel to the Go to page (#766) and Go to line (#767) PRs. Five commits. Unlike those two, the 0-100 slider stays — it makes sense for percentage selection even though it doesn't for pages or lines.

## 1. Land initial focus on the numeric entry

The dialog opened with focus on the slider, so a screen-reader user had to tab past it to reach the percentage field every time. Focus now lands on the entry so a percentage can be typed directly. Tab order is unchanged (it follows widget creation order), so the slider remains the first tab stop, matching on-screen order.

## 2. Interrupt the focus-chain announcement

Closing Go to percent made NVDA announce the whole focus chain ("Paperback, tab control, ...") before the jump was spoken. The shared delay-announce helper (`announce_after_delay` in `navigation.rs`, the same one the other two PRs add) raises the announcement ~30ms after the dialog closes, cutting off all but a sliver of the chain.

The announcement reads **the line the jump lands on**, as Go to Line does — a percentage alone says nothing about where it points. A blank line falls back to the percentage itself ("8%").

## 3. Reject out-of-range percentages

Typing a percentage outside 0-100 was silently clamped by the spin control, so a huge number jumped to 100%. The entry is now a plain text field that validates on submit: out-of-range or unparseable input announces "Percent out of range." and keeps the dialog open, mirroring the other Go dialogs. The slider is kept and stays in step with the field both ways (dragging rewrites the field; typing moves the slider), and the label now reads "Go to percent (45/100):".

## 4. Accept +n and -n

For consistency with Go to line and Go to page, `+n` moves n percentage points forward and `-n` back from the current percentage ("+10" from 45% is 55%). A bare number still means the percentage itself. Relative results are held to the same 0-100 rejection, and the slider follows the resolved value as it's typed. Unit tests cover the parser.

## 5. Block non-numeric keystrokes

A `CHAR`-event filter swallows anything that can't be part of a percentage expression — digits, `+`/`-`, and Enter — so letters can't be typed. Only character-insertion events are filtered, so control codes (Tab, Backspace, Enter, Escape), Delete, the special keys at `WXK_START` and above (arrows, Home/End, Page keys, Insert, F-keys, numpad Enter), and all Ctrl/Cmd shortcuts keep working.

## Notes / tradeoffs

- `announce_after_delay` is added to `navigation.rs` here exactly as in #766 and #767; all three PRs carry the identical addition, which git auto-merges when the first one lands.
- The entry switched from a `SpinCtrl` to a plain text field (nothing visually changes) — that's what makes rejection, relative input, and the keystroke filter possible.
- New translatable strings: "Go to percent (%d/%d):", "Percent out of range." (removed the old "P&ercent:"). Pot regeneration left to the usual maintenance pass.
- Only NVDA is verified, same caveat as the other PRs.

## Testing

Tested with NVDA on Windows 11: Ctrl+Shift+G lands in the entry and reads the line you jump to with no focus chain after it; out-of-range and garbage input speak "Percent out of range." and stay in the dialog; `+10`/`-5` land where expected; letters can't be typed while Tab, arrows, Backspace and shortcuts all still work; the slider still controls the value and tab order is unchanged; ESC/Cancel unchanged; the four parser unit tests pass.

### DIFF
--- a/crates/paperback/src/ui/dialogs/go_to_percent.rs
+++ b/crates/paperback/src/ui/dialogs/go_to_percent.rs
@@ -74,6 +74,9 @@ pub fn show_go_to_percent_dialog(parent: &Frame, current_percent: i32) -> Option
 	add_ok_cancel_footer(main_sizer, ok_button, cancel_button);
 	dialog.set_sizer_and_fit(main_sizer, true);
 	dialog.centre();
-	percent_slider.set_focus();
+	// Focus the numeric entry so a screen-reader user can type a percentage directly.
+	// Tab order is untouched (it follows widget creation order), so the slider remains the
+	// first tab stop, matching the order the controls appear on screen.
+	input_ctrl.set_focus();
 	if dialog.show_modal() == ID_OK { Some(input_ctrl.value().clamp(0, 100)) } else { None }
 }
--- a/crates/paperback/src/ui/dialogs/go_to_percent.rs
+++ b/crates/paperback/src/ui/dialogs/go_to_percent.rs
@@ -1,9 +1,9 @@
-use std::{cell::Cell, rc::Rc};
+use std::rc::Rc;
 
 use patois::t;
 use wxdragon::prelude::*;
 
-use super::DIALOG_PADDING;
+use super::go_to::{NumberEntry, add_go_cancel_footer};
 
 pub fn show_go_to_percent_dialog(parent: &Frame, current_percent: i32, live_region_label: StaticText) -> Option<i32> {
 	let current_percent = current_percent.clamp(0, 100);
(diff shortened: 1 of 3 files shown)


johanntan:
Hi,
Based on reading this, I personally think that focus on the slider on opening the dialog makes more sence, because like adjusting a percent of something, like volume for example has always been done with sliders.
This is just my opinion of course and I would like to see other's thought's on this.

dan1982code:
Ah, thank you for discussion! I did that for parallelism with how go to page and go to line work... edit box first so you can quickly type a percentage instead of arrowing a ton. But I don't have a super strong opinion for me personally on that :)

trypsynth:
Agreed with @johanntan, the keyboard focus should be on the slider by default I think.

dan1982code:
On board :) BTW, I know that the + and - relative movement isn't super useful for percents. But it is useful for pages and lines, so I kept it here for percents too just for consistency.

Issue #743, opened 2026-08-31
Android issues


piano777wzy:
Hello, below are the issues I encountered while using the Android version.
First, some translation issues:
On the first launch, there is a start screen with a sentence that cannot be translated:
"Powers the optional in-app file browser, so you can open documents from anywhere on your device — including network drives — instantly, with full screen-reader support. You can skip this and use the system file picker instead."
Also, I don't know why, but whenever I open a new screen, my screen reader first reads the English text and then the Simplified Chinese translation I made. For example, when I find "Find" which is "查找" in Chinese, after tapping it, it first reads the English "Find" and then the Chinese "Find". I don't know what's going on. The same happens on other screens. For example, in the Table of Contents, when I go back from it or open it, there is an English "Table of Contents" announcement, followed by the Chinese "Table of Contents". This is very strange.
Also, some elements like buttons, headings, checkboxes, etc. cannot be translated at all. For example, "Heading", "level", and checkboxes such as the "not ticked" state before "Use regular expressions" in the Find screen—these element announcements cannot be translated.
Another issue: My Table of Contents is read twice. No matter which ebook I open, every entry in the Table of Contents is read twice. This should happen with any file.
Issue: In TTS mode, when I open the Table of Contents and jump to a section, there is a problem. It immediately goes to the beginning of the book. It reads the title of that section once, and then immediately jumps to the beginning of the file.
That's all for now.

trypsynth:
Taking these one at a time, since they are four different things.

**The onboarding sentence.** This one is translated, and has been since 9b42aaf5 on 29 August: it is in `po/zh_CN.po` and in the catalog the app ships, `android/app/src/main/assets/translations/zh_CN.json`, and the screen reads it through `t()` like everything else. So on a build newer than that it should be Chinese. If you are still getting English, tell me the version you are on and I will look again, because then it is the lookup failing on that one entry and not a missing translation.

**Elements that cannot be translated.** You were right, and these are now fixed. Fourteen strings were written straight into the TalkBack descriptions in English instead of going through `t()`, so no locale could ever translate them:

- "Heading 2" on a heading line
- ", Level 2" on a table of contents entry and on an elements list heading
- "50 percent" on the Go to slider, and on the text size, speech rate and pitch sliders
- "Current path: ..." above the file browser
- "myfile.epub, File, modified ..." on a file browser entry
- the action labels TalkBack reads as "double tap to ...": "go to chapter", "go to heading", "go to link", "open", "show import and export options"

[...]

PR #806, opened 2026-09-07
Android: keep the position a table of contents jump moved to, part two

### LINKED ISSUE #794: Issue: On Android, tapping a chapter in the table of contents while in TTS mode still jumps to the beginning of the book.
Hello, on the Android version, in TTS mode, if playback is active, when I open the table of contents, find a section, and tap it, after a short while it jumps directly to the beginning of the book. That is, after reading the chapter title and then a couple of sentences, it suddenly jumps straight to the beginning. This issue still exists.

trypsynth:
Closes #794.

Picking a chapter in the table of contents while read-aloud was running still sent the reader back to the start of the book a moment later. #743 fixed one of the two paths that did this; here is the other.

The reading position is also saved from the text list's scroll offset:

```kotlin
snapshotFlow { listState.firstVisibleItemIndex }.distinctUntilChanged().debounce(500)
```

`snapshotFlow` reports where the list already is as soon as anything collects it. Opening the table of contents pops MainScreen off the back stack, so coming back composes it afresh and builds a new `LazyListState` at `initialScrollIndex`, the line the document was opened at. That opening index is the flow's first value, and half a second later `savePosition` writes it over the section just chosen, in `_ttsPosition` and in the config both. The title and the paragraph after it were already queued with the text-to-speech engine, so they play before the jump shows up, which is why it looks like the book reads on for a sentence or two and then starts over.

The first value a list reports is never a place the reader went, so it is not saved. Everything after it is a real scroll.

### DIFF
--- a/android/app/src/main/kotlin/dev/paperback/android/ui/ListScrollPosition.kt
+++ b/android/app/src/main/kotlin/dev/paperback/android/ui/ListScrollPosition.kt
@@ -1,0 +1,23 @@
+package dev.paperback.android.ui
+
+import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.debounce
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.drop
+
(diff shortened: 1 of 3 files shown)


trypsynth:
Pushed straight to master.

PR #848, opened 2026-09-14
Polish: TalkBack labels, reading position, RAR notice, help updates

michaldziwisz:
Translates the twelve strings added by the Android accessibility work and the recent parser changes.

### TalkBack action labels use the infinitive, not the imperative

Compose reads `onClickLabel` out as "double tap to *label*", so these are completions of that sentence rather than commands. Polish uses an infinitive in that position; an imperative would give the equivalent of "double tap to go-you-to heading".

| string | Polish |
|---|---|
| `go to heading` | przejść do nagłówka |
| `go to link` | przejść do odnośnika |
| `go to chapter` | przejść do rozdziału |
| `open` | otworzyć |
| `show import and export options` | pokazać opcje importu i eksportu |

### State and content descriptions

| string | Polish |
|---|---|
| `Heading {}` | Nagłówek {} |
| `{} percent` | {} procent |
| `{}, Level {}` | {}, poziom {} |
| `Current path: {}` | Bieżąca ścieżka: {} |
| `{}, {}, modified {}{}` | {}, {}, zmodyfikowano {}{} |

### Reading position and the RAR notice

`%s, page %d` → **%s, strona %d**. `position_announcement` already formats the percent sign into `%s`, so the Polish carries no separate word for "percent" — it reads "15%, strona 30".

The RAR message is translated in full, keeping the instruction to repack as `.cbz`.

`Link` is rendered **odnośnik**, matching the eleven entries that already use it.

### All eight fuzzy matches were wrong, not approximate

`msgmerge` filled them from unrelated entries, so all eight were rewritten from scratch. The clearest cases:

* `go to link` came from `Go to line` — a different destination entirely.
* `go to chapter` came from `Go to percent`.
* `Current path: {}` came from `Currently Open`, losing the path.
* `{} percent` came from `Percent`, dropping the number the slider is set to — TalkBack would have announced a bare "Procent" with no value.

### One msgid disappeared, and that is correct

The password-protected PPT notice is gone. It is absent from both `paperback.pot` and the source tree, because 78016f35 implemented support for those files, so the message saying they are unsupported no longer has anything to report. It stays in the catalog as an obsolete `#~` entry rather than being deleted.

### Help file

`doc/readme-pl.md` was missing the `.cbr` extension, the manual pages entry, and two 0.9.2 changelog bullets (the `=` shortcut reporting the page number, and the relaunched window coming to the front after an update). Added in the order the English original uses.

### Verification

* `msgfmt --check --statistics`: 910 translated, 0 fuzzy, 0 untranslated.
* Plural forms checked by **running** `ngettext` against the compiled `.mo`, not by reading the `Plural-Forms` rule, with its own validity control.
* Help structure against English: 66/66 headings, 118/118 backticked items.
* Key-translation pairs between `HEAD` and the working tree: **12 msgids added, 1 removed, 0 existing translations altered.** The diffstat on `po/pl.po` is larger than that because `msgmerge` rewraps lines and refreshes the header date.

Not verified: I have no Android device with a screen reader here, so the infinitive form follows from how Compose composes the announcement, not from listening to TalkBack say it.


michaldziwisz:
`build-macos` failed here on `hdiutil: create failed - Resource busy`, which is the DMG packaging step, not the build. The same log shows `Built app: .../Paperback.app` a few lines earlier, and `Tests (macOS)` passed on the same commit.

I could not rerun the job in this repository, so I ran `build.yml` on the fork from the identical commit (`2fdb871a`): **build-macos succeeded there**, along with the other five targets. Same code, same workflow, different runner — so this looks like a transient runner condition rather than anything in the change.

Worth noting either way: this PR only touches `po/pl.po` and `doc/readme-pl.md`, so it has no path to the packaging step.

PR #850, opened 2026-09-14
Add MathML formula support

mikolysz:
- **feat(math): set up core math infrastructure**
  This is the first in a series of commits adding math support to Paperback.
  It adds the tedious plumbing necessary for the rest of the math
  implementation to function:
  
  * a Formula marker in `document/marker.rs` and its UDL declaration,
  * a `FormulaInfo` type, similar to the other `*Info` types already present,
  * formula navigation and a Formula reading unit through `SegmentTypeFfi`,
  * formula markup lookup at the current reading position,
  * tests for marker bounds, navigation and supported reading units.
  
  The design of this implementation is loosely modeled on how tables are
  implemented. Table and formula lookup share the same checks for whether
  the current position falls inside a marker's display span.
  
  Handle `SegmentTypeFfi.FORMULA` in `NavUnit.kt` and display it as
  Formula in the reading unit picker. This belongs with the enum
  addition because the Kotlin `when` must remain exhaustive after
  regenerating bindings. Include a translator comment for the Formula
  label.
  
  Android gets its supported reading units from the shared core, and
  its previous/next controls already use the shared segment navigation.
  The parser commit supplies the formula markers those controls need.
  
  Android has no M/Shift+M formula shortcuts or Formula View yet. Both
  should be implemented and tested with someone who has access to an
  Android testing setup. Formula View also needs the existing markup
  lookup exposed through UniFFI and an Android UI.
  
  This is the first commit in the PR that introduces a translatable math
  UI string: the Android Formula label.
  
  POT regeneration is intentionally deferred for the entire PR:
  the pinned patois-build 0.1.13 chooses mobile translator comments
  according to unsorted filesystem order, so it can replace notes without
  source changes. Running cargo gen-pot also picks up unrelated changes
  from the existing upstream sources, including the reordered Results
  entry. We are not checking those unrelated changes into this PR.
  
  Regenerate po/paperback.pot once a fixed patois-build version is in use
  and the unrelated baseline refresh has been handled separately. The
  deferred regeneration must include the Formula label introduced here
  and the math UI strings in the later desktop commit.
  
  On macOS, all five formula session tests pass in release mode with
  UniFFI enabled, and core formatting passes. Kotlin and Swift bindings
  generate successfully from the UDL introduced here.
  
  This change has not been built or tested on Android.
  

- **feat(math): add math parsing**
  This commit adds math support to Paperback's parsing machinery (notably
  `OutputConverter`) and wires MathML conversion into `html_to_text` as
  well as `xml_to_text`. EPUB, HTML and CHM use these shared converters.
  
  Conversion from MathML to AsciiMath happens via MathCAT, which, since
  it is also written in Rust, is added as a normal Rust crate via Cargo.
  We enable the feature that embeds its rules directly into the binary,
  so no separate rules directory needs to be shipped. We convert to
  AsciiMath, a screen-reader friendly, language-neutral, lightweight text
  notation exposed through MathCAT's braille API. MathCAT can also
  convert to speech in English and a few other languages, as well as
  Braille codes like Nemeth or UEB. Such output preferences are left for
  a later change, and are not included in this patchset.
  
  The math parser implemented here contains a formula memoization cache
  shared across threads. As books tend to repeat expressions frequently,
  sections can reuse earlier conversions. The cache caps stored keys and
  rendered text at 8 MiB. MathCAT initializes lazily on each worker in
  the existing parallel EPUB parser. This adds ~100ms of unavoidable
  parsing time to books containing math content. However, as Rayon tends
  to use a shared thread pool for its workers, this initialization likely
  won't be repeated across different books.
  
  We keep the MathML and display span of each formula reached by the
  normal document walk for navigation and Formula View. If conversion
  fails, we fall back to the author's alternative text, then the
  element's text content.
  
  XML links and table cells flatten formulas to raw token text. HTML
  links still convert formulas during their normal tree walk, but links
  and tables have no formula markers: navigation skips those formulas and
  Formula View cannot open them. Separately collected navigation labels
  and captions also use raw text, even where the reading buffer contains
  AsciiMath.
  
  This follows an existing bug in Paperback that relates to how content
  within links and headings is handled. General handling of nested
  content, including formatting, images and anchors, would necessitate
  too much unrelated work and belongs in a separate parser change. XML
  tables still pass through source slicing and HTML reparsing, so
  preserving inherited XML context there is deferred with that work.
  
  A small XML adapter emits standalone, unprefixed MathML for MathCAT
  and Formula View, which loads HTML. It uses the parsed tree's resolved
  namespace bindings and entity values, keeping prefixed formulas and
  document-defined entities usable outside their source document. It
  preserves ordinary attributes and standard xml: attributes. Foreign XML
  subtrees and custom namespaced attributes are omitted; general foreign
  XML preservation is deferred. The HTML converter uses its existing
  element serializer directly; both consumers accept its named entities.
  
  Recognition accepts MathML and unnamespaced XML math elements; elements
  in other namespaces remain ordinary content in the normal XML tree
  walk. HTML recognition uses the namespace assigned by the HTML parser.
  
  We add HTML and XML regression cases for embedded formula text, spans
  and anchors across ordinary containers. Further cases cover inline/block
  display, fallback text, namespace recognition, inherited prefixes and
  entities. Formula View checks parse the retained payload as HTML and
  assert its MathML structure, text and relevant attributes. These check
  the view's input, not native WebView rendering.
  
  Deferred cases keep their intended assertions under
  `#[ignore = "deferred: ..."]` for links, tables, navigation labels and
  foreign annotations. TODOs identify cases to merge when shared rendering
  handles nested content and labels, and when foreign annotations can
  reach the HTML view.
  
  The IDPF linear-algebra EPUB used for the measurements below is also
  available for manual validation. Checks against this external book
  belong outside the repository's test suite.
  
  This parser is unable to distinguish between block and inline math
  content if such distinctions are signalled via CSS only. This is not
  supposed to happen (and in the documents I've seen, it doesn't happen;
  the relevant attribute on `math` is always present), but I have a
  suspicion that other documents might not be so conformant. There is
  probably very little we can do about this if we don't want to add a
  proper CSS parser to Paperback.
  
  Measurements on September 14, 2026, using an Apple M1 with 8 CPU cores
  and 16 GiB RAM, macOS 15.3 and rustc 1.96.1:
  
  * Build pb at this parser revision with
    `cargo build --release --locked -p pb`, using the repository's stripped
    release profile (opt-level=z, LTO and one codegen unit).
  * Convert the IDPF sample's 2023-07-04 release with
    `RAYON_NUM_THREADS=8 pb linear-algebra.epub --no-prompt -o output.txt`.
    The pinned download is
    https://github.com/IDPF/epub3-samples/releases/download/20230704/linear-algebra.epub
  * Five fresh processes after one warmup take a median 3.31 seconds
    (range 3.20-3.47 seconds), including text-file export. The filesystem
    cache is warm; each process starts with an empty formula cache.
  * This revision's pb binary is 5,356,080 bytes (5.11 MiB), compared with
    3,494,800 bytes (3.33 MiB) at the pre-math base 2987656f6466:
    an increase of 1.78 MiB, or 53.3%.
  
  These timings cover this branch's fixed AsciiMath output and the 9,776
  formulas reached by the normal document walk. Math preferences and
  per-language speech output are still to come, so this branch is much
  faster than the speech mode planned for a later change.
  Expect substantially longer conversion times when speech output is
  enabled.
  
  For the same book's formulas, a separate one-thread MathCAT 0.7.5
  benchmark with an 8 MiB cache took 11.8-12.0 seconds for AsciiMath and
  66.3-66.9 seconds for English speech (two runs each), about a 5.6x
  difference. Speech builds an intermediate interpretation tree before
  applying language-specific rules.
  
  On macOS, all 169 active converter tests pass in release mode at this
  revision. All 24 deferred converter cases fail on their documented
  limitations when run explicitly. Core formatting and core/pb Clippy
  with UniFFI enabled pass with the repository's lint settings; Clippy
  reports no warnings in files changed by this commit. Every benchmark
  process produces the same export; the deferred table rendering uses raw
  formula text.
  

- **feat(math): add UI support for math**
  This commit wires up the math handling infrastructure into the
  wxWidgets-based desktop UI. It implements:
  
  - a Formulas submenu with commands to go to the next and previous
    formula, using M and Shift+M by default and supporting custom
    keyboard shortcuts,
  - a way to view a particular formula's MathML in a web view with Enter
    or Space,
  - translatable strings and translator notes for menu items, shortcut
    actions, navigation announcements and Formula View.
  
  We release the DocumentManager lock before opening the modal dialog,
  whose handlers may need to acquire it again. Formula View is available
  for formulas reached by the normal document walk. Links and table cells
  have no formula markers. XML links, table cells and navigation labels
  still flatten formulas to raw text.
  
  Tests ensure that formula shortcuts are discoverable, can be
  rebound, and survive a configuration round trip.
  
  POT regeneration is deferred for this PR as explained in the
  infrastructure commit.
  

- **docs: mention math support in readme**
  Document MathML-to-AsciiMath conversion, formula navigation and Formula
  View in the README and user help, including the navigation and
  activation shortcuts.


trypsynth:
Thanks for this! Just started reviewing. However:

> POT regeneration is intentionally deferred for the entire PR:
> the pinned patois-build 0.1.13 chooses mobile translator comments
> according to unsorted filesystem order, so it can replace notes without
> source changes. Running cargo gen-pot also picks up unrelated changes
> from the existing upstream sources, including the reordered Results
> entry. We are not checking those unrelated changes into this PR.
> Regenerate po/paperback.pot once a fixed patois-build version is in use
> and the unrelated baseline refresh has been handled separately. The


What are you talking about? This sounds like a real bug in patois-build, please submit a pr for it or an issue, I am currently very confused by what you mean. I can gen-pot and it works perfectly, nothing about...mobile translator comments getting reordered. Mind explaining?

mikolysz:
@trypsynth From [`std::fs::read_dir` docs](https://doc.rust-lang.org/std/fs/fn.read_dir.html)

> The order in which read_dir returns entries can change between calls. If reproducible ordering is required, the entries should be explicitly sorted.

And they aren't in Patois. When I was filing this, I ran into this race condition (+ some bad pot on old master which is no longer a problem after rebasing) which made the pot diff balloon with non-math changes. This probably depends on something as simple as my OS, or just which order some parallel thread pool happened to write the repo files when cloning or something.

I'll indeed file something for this upstream though.

trypsynth:
Ah, gotcha. Okay, that took me off guard and as such I admittedly didn't review the rest of the pr body as closely as I should've because I jumped straight to a bug in patois-build. Thanks, will review parts of this, chat on the issues etc. This will be epic!

mikolysz:
The patois bugfix is now pr#2 on that repo btw. The bug is definitely there (it relies on a function which doesn't return results in a deterministic order according to Rust docs), but somehow the issue hasn't happened to any of us before, despite me always being on the Mac and you mostly being on Windows. I can't really reproduce it any more either, but the issue is definitely there (and not an AI hallucination because I worked on the diff and saw the reorder myself, which is why I removed the quite-large pot diff from being checked in by git in the first place). The best explanation I can come up with is that I was rebasing the old version of this PR in a Chat GPT work environment, which runs Linux, and the filesystem there orders files differently somehow. This is honestly quite strange, and while this bug is definitely a bug, I don't know if it's the only bug.

trypsynth:
Ran this on Android (Pixel 9a, debug build off the rebased branch) since the PR notes it had never been built or tested there. The desktop/core side is solid, but Android has a blocking bug in the formula text itself.

## Formula text comes out as Nemeth braille on Android, not AsciiMath

Opening a book with formulas, selecting the **Formula** reading unit (which does show up correctly), and stepping to a formula, the reader lands on Unicode braille cells:

```
⠭⠀⠨⠅⠀⠹⠤⠃⠬⠤⠜⠃⠘⠆⠐⠤⠲⠁⠉⠻⠌⠆⠁⠼
```

That is Nemeth (`⠨⠅` is the Nemeth `=`). The exact same EPUB through `pb` on desktop gives the intended AsciiMath:

```
x = (-b+-sqrt(b^2-4ac))/(2a)
```

So on Android the feature emits braille dot patterns, which TTS can't read as math. Same PR code, same input, same embedded MathCAT rules on both platforms.

## Root cause

`init()` sets `BrailleCode = ASCIIMath` per worker thread, and MathCAT's preference manager is thread-local, so that should reach `get_braille` on the same thread. But `braille::braille_mathml` calls `read_files()` on every `get_braille`, and `read_files()` re-runs `set_preference_files()` unless `CheckRuleFiles` is `"None"` (mathcat 0.7.5, `speech.rs:2215`). That reload resets `BrailleCode` to its shipped default, Nemeth, discarding the programmatic ASCIIMath before the braille is generated.

I confirmed this directly by logging the effective `BrailleCode` right before `get_braille` on the device: it read `Nemeth`, despite `init()` having set ASCIIMath and returning success. Desktop doesn't surface the same reset; Android does.

## Fix (verified on device)

Set `CheckRuleFiles` to `None` in `init()` so MathCAT stops reloading its preference files and the ASCIIMath setting sticks:

```rust
let result = libmathcat::set_rules_dir("Rules".to_string())
    // Without this, MathCAT re-reads its preference files on every get_braille call, which
    // resets BrailleCode to its shipped default (Nemeth) and discards the ASCIIMath set below,
    // so the reader hears Nemeth braille cells instead of AsciiMath.
    .and_then(|()| libmathcat::set_preference("CheckRuleFiles".to_string(), "None".to_string()))
    .and_then(|()| libmathcat::set_preference("Language".to_string(), "en".to_string()))
    .and_then(|()| libmathcat::set_preference("BrailleCode".to_string(), "ASCIIMath".to_string()));
```

With that in, the same navigation on the same device gives `e^(i pi)+1 = 0`, desktop still gives AsciiMath, and `cargo test -p paperback-core` (82 formula tests) and `cargo +nightly fmt --check` stay green.

## Other notes

- iOS: won't break and won't surface the feature. The iOS app hardcodes its reading units as a 4-case Swift enum (`SegmentType` in `AppViewModel.swift`) rather than reading `supported_segment_types()` like Android, so no structural unit (table/image/formula) is exposed there today. Pre-existing, out of scope here.
- Cosmetic: "MathCat" typo in the doc comment at `formula.rs:110` (should be "MathCAT").

PR #851, opened 2026-09-14
Mark where a selection begins, then copy everything up to where you got to

dan1982code:
## Mark where a selection begins, then copy everything up to where you got to

Selecting a lot of text is the awkward part of reading with a screen reader. Shift+arrow takes text a
line at a time, and shift-click needs a mouse, so "everything from here to there" has no affordable
keyboard answer. This is a **proof of concept** for one:

* **Alt+F9** marks the beginning of a selection and announces *"Beginning of selection set."*
* **Alt+F10** copies everything from the mark to the current position and announces
  *"Copied (x chars)."*
* **Alt+Shift+F9** jumps back to the mark, leaving it set, and announces
  *"Jumped to beginning of selection."*
* Either key with nothing marked announces *"Cannot copy; set beginning of selection first."* /
  *"Cannot jump; set beginning of selection first."*

All three are rebindable in the existing Customize Keyboard Shortcuts dialog, and all three sit under
a **Tools > Select and copy** submenu. The distance between the presses is unlimited — mark near the
front of a book, read on for a few hundred pages, copy the lot, then jump back to where you started
from.

**I am posting this to find out whether the idea is worth pursuing, not as something ready to merge.**
See *Testing* for what has and has not been verified.

### Why Alt, and not Shift+F9 / Shift+F10

The first sketch of this used Shift+F9 and Shift+F10, which cannot work: **Shift+F10 is already the
reader context menu** (`ui/reader_input.rs:108`, documented at `doc/readme.md:153`). On Windows
Shift+F10 *is* the context-menu key, so rebinding it would take the context menu away from anyone
without a Menu/Application key — the readers this feature is for. Alt+F9 and Alt+F10 are unbound
today.

Alt chords are a proven route rather than a hopeful one: `ActionId::GoBack` is an ordinary entry in
the command table whose default chord is Alt+Left, and it ships. Alt chords reach their commands
through wxWidgets' menu accelerator, parsed from the label's `\tAlt+Left` suffix.

### Behavior

* The mark is **cleared by a copy**, so a stray second Alt+F10 says "cannot copy" rather than quietly
  taking half a book. Alt+F9 moves the mark at any time.
* Pressing Alt+F10 **above** the mark copies the range between the two in document order, the way
  dragging a selection upward behaves, rather than refusing.
* **Both ends are included.** A caret names the character it sits *on* — that is the character a
  screen reader is speaking — so a range that ended at the caret would drop the last character the
  reader stopped on deliberately. The end is therefore reached past that character
  (`display_pos_after_char_at`), spanning the whole of it where it is an astral-plane pair. This
  differs from a mouse drag, which selects up to but not including the character under the cursor,
  and it is the difference the first test build caught.
* A copy that comes out empty — marking and copying with nothing under the caret, i.e. an empty
  document — announces *"Nothing to copy."* and **leaves the clipboard alone**; replacing what
  someone had copied with an empty string, to announce "0 chars", would destroy data for nothing.
  This is a fourth message beyond the three above.
* The count is **characters**, not the display-unit width of the range: a spoken count is for someone
  listening, and an astral-plane character is two display units but one character to a person.
* The visible selection is **deliberately not set**. The range can be larger than the loaded window,
  and `to_local` clamps rather than fails, so the control would show a selection that is quietly not
  what was copied. It would also disturb commands that read the control's selection: Word Count would
  start reporting the selection instead of the document, and Toggle Bookmark would bookmark the whole
  range. The announcement is the feedback; the mark is a remembered position, not a highlight.

### The part that would have corrupted text silently

Positions in this app are **display units**, not chars — the caret, `Marker.position` and
`TextWindow` bounds all use them, and `total_display_len()` is UTF-16 code units on Windows and macOS.
`DocumentSession::get_text_range`, the obvious call, is **char**-indexed: it slices via
`byte_index_for_char` and clamps to `char_count()`.

For an interior range those two units disagree wherever a document holds an astral-plane character,
and the text would come back shifted by however many preceded it — a wrong copy that still announces
a plausible character count. The buffer's own test makes it concrete (`document/buffer.rs:537-557`):
for `"a"` plus an emoji, display offset 3 is byte 5.

The whole-document copy that exists today (`copy_whole_document_if_all_selected`) only survives this
because it asks for `(0, doc_len)` with `doc_len` in display units, and clamping to the char count
happens to land exactly on the end. So this adds `get_text_range_display`, which converts through
`byte_index_for_display` and is placed in `session/window.rs` rather than next to `get_text_range` —
the file that owns "display-unit position" as a concept, so the two are not sitting side by side
inviting the mistake.

### What changed

* **paperback-core**
  * `DocumentSession::get_text_range_display` — a display-unit-indexed text range, plus tests
    covering a mid-word interior range, clamping, and an astral-plane case asserting both the correct
    result and that the char-indexed sibling disagrees on the same numbers.
  * `DocumentSession::display_pos_after_char_at` — the inclusive end bound: the position just past
    the character a caret is sitting on, so the copy covers the last character read and spans a whole
    astral-plane pair rather than half of one.
  * `ActionId::SetSelectionStart` / `CopyFromSelectionStart` / `JumpToSelectionStart`, in the Tools
    category, defaulting to Alt+F9 / Alt+F10 / Alt+Shift+F9 (Cmd+F9 / Cmd+F10 / Cmd+Shift+F9 on
    macOS — see below). A test pins the chords to the key codes the text control actually reports and
    asserts that none of the three answers for another's chord, so a typo in a key string fails a
    test rather than shipping a shortcut that never fires.
* **paperback**
  * `ui/selection.rs` — the copy and its announcements. The document is read under the lock, which is
    then **dropped before the clipboard write**: `Clipboard::set_text` goes through OLE on Windows,
    which can pump messages, and the document manager's mutex is not reentrant.
  * A `selection_mark` on `DocumentTab` (per document, so a mark cannot leak between tabs), with the
    two places the buffer moves in place handled: `apply_ocr_results` **shifts** the mark the same way
    it shifts the caret, and `reparse_tab_in_place` **clears** it, because the caret there is
    re-derived from an anchor or a percentage rather than carried over.
  * Three `COMMANDS` rows and a **Select and copy** submenu in the Tools menu, built from
    `commands::menu_entries` + `build_menu` like the Go menu's groups. The jump reuses
    `move_to_offset_and_record_history`, so Alt+Left comes back to wherever you jumped from, the way
    it does after a bookmark jump.
  * New translatable strings: the three announcements, the two unset-mark messages, and the
    submenu's label and help text.
  * `reader_input.rs` dispatches these two explicitly even when Alt is held. Alt chords are normally
    left to the menu accelerator, but F10 means "activate the menu bar" to Windows, and whether
    Alt+F10 reaches the accelerator table on every platform is not something to bet a default on. The
    two routes cannot both run — a keystroke the accelerator consumed never reaches the control.

### Testing

Automated, all green: `cargo test -p paperback-core` (946) and `cargo test -p paperback --bins` (97),
`cargo clippy --workspace --all-targets` adds no new warnings, and `cargo +nightly fmt --check` is
clean.

**A release build has now been run on Windows and the feature works**, so the keyboard handling —
including the Alt+F10 question above, which was the main risk — is settled: both shortcuts fire and
the copy lands on the clipboard. The first build exposed the inclusive-end bug described under
*Behavior* (the last character of every selection was missing), which is fixed here and covered by
tests at both ends.

Still unverified, and what I would want to see before this went anywhere:

* **Exactly one dispatch per press.** The command is reachable both through the menu accelerator and
  through the explicit dispatch in `reader_input.rs`, and they are meant to be mutually exclusive. A
  `tracing::debug!` in each handler should print one line per press, not two.
* A copy across a document large enough to have its loaded window swapped out between the two presses
  (the repo's large PDFs are handy for this).
* A copy where an astral-plane character falls before the marked position, pasted somewhere it can be
  compared with the document.
* The two invalidation paths: a Batch OCR that changes the page count, and a file touched on disk to
  trigger the reparse.
* **The jump** (Alt+Shift+F9) has not been run at all — it was added after the first round of
  testing. In particular, whether the caret lands where the mark is across a window reload, and
  whether Alt+Left really returns to where it was pressed from.

**macOS is untested and the defaults there are a guess.** Option+F9/F10 is a media-key chord on an
Apple keyboard and the system consumes it before the app sees it, so the macOS arm uses Cmd+F9/Cmd+F10
instead — the same reason `GoBack` is Alt+Left on Windows and Cmd+`[` on macOS. That needs checking on
a real Mac, or a different default.

### Not included

* Any visual indication of the mark. Deliberate, for the reasons in *Behavior*.
* Copying formatting to the clipboard. This is plain text, like the existing whole-document copy.
* Persisting the mark across restarts, or across documents. It is per-tab, in memory, and a mark
  names a position in one document.


dan1982code:
Pushed a fix for a bug found on the first test build: **the last character of every selection was missing.**

The range was end-exclusive (`[mark, caret)`). That is what a mouse drag does, and it is the wrong model here — a caret names the character it sits *on*, which is the character a screen reader is speaking, so stopping on the last character you want dropped exactly that one.

Both ends are inclusive now. `display_pos_after_char_at` gives the end bound that reaches past the character under the caret, spanning the whole of it where it is an astral-plane pair rather than half of one. `get_text_range_display` keeps its honest exclusive bound; the conversion happens at the call site.

This also means "Nothing to copy." now only appears on an empty document, rather than whenever the two presses landed in the same place.

That build also confirmed the part I was least sure of: **both shortcuts fire and the Alt+F10 question is settled.** I have updated the description's *Testing* section — it no longer claims nothing has been run, and what remains unverified is down to a cross-window copy, an astral character before the mark, the two buffer-mutation paths, and a check that a press dispatches exactly once (it is reachable both through the menu accelerator and the explicit dispatch in `reader_input.rs`, which are meant to be mutually exclusive).

dan1982code:
Added a third command and grouped the set, at the request of the first round of testing:

* The three now live under **Tools > Select and copy** instead of sitting loose in the Tools menu.
* **Alt+Shift+F9 — Jump to Selection Start**, which puts the reader back at the mark without consuming it, so mark / jump / copy can be used in any order and as often as you like. It routes through `move_to_offset_and_record_history` like every other jump in the app, so Alt+Left returns to wherever it was pressed from, and it handles the window reload itself when the mark is further away than the loaded window covers.

The chord test now checks all three against each other, so Alt+F9 and Alt+Shift+F9 cannot start answering for one another.

The two earlier commands are confirmed working from that testing. **The jump has not been run yet** — it went in afterwards — so a caret landing on the mark across a window reload is the part to watch.

trypsynth:
Holding this for now rather than merging, on the strength of your own "not ready to merge" and the unverified list. The idea is worth pursuing and the copy path looks solid; two things would move it over the line:

- **The jump (Alt+Shift+F9) run for real.** It has never been exercised. Mainly whether the caret lands on the mark across a window reload, and whether Alt+Left then returns to where the jump was pressed.
- **macOS.** The Cmd+F9/F10 defaults are a guess; they need a real Mac, or a default you have confidence in.

The single-dispatch check and the large-window / astral-before-mark / OCR-and-reparse invalidation cases would be good to close out too, but the two above are the blockers. Re-request merge once the jump and macOS are confirmed and I will take another look.

trypsynth:
Follow-up after reading the jump path closely: the logic is sound, so the earlier worry about it is mostly settled on paper.

The jump is three calls: capture the current caret, `set_caret_to_doc_offset(tab, mark)`, then `record_history`. That is the same sequence bookmark-jump uses, and `set_caret_to_doc_offset` is the single chokepoint every caret jump in the app routes through (bookmarks, go-to-line, Find) which reloads the window around an off-screen target via `needs_reload_for`. So "the caret lands on the mark across a window reload" and "Alt+Left returns to where the jump was pressed" both ride on the shipping mechanism, not on new code. The branch builds and the full suite is green.

What is left is only what a headless check cannot reach: pressing Alt+Shift+F9 in the running app and confirming it fires, announces "Jumped to beginning of selection.", and that Alt+Left then returns. Quin can run that, and macOS too, so the Cmd+F9/F10 defaults can be confirmed rather than guessed. Once those are in, this is good to merge.

dan1982code:
I'm testing here, working well so far. One thing, when i jump to the start of selection, maybe it should read the line it lands on rather than just telling me it jumped to the start? I'm not sure on this. Usually I want to know what line I land on, but for some reason here I want to be confirmed that I jumped to the start of selection. Might just be me.

trypsynth:
No, I'd agree, it should announce that and then the line you land on, I think.

trypsynth:
@dan1982code is this pr working as you want? I'll review it after work today if so.

dan1982code:
@trypsynth yes, but community testing/preferences on this one would be good to have. Hoping for feedback if PR is merged. Thanks!

PR #902, opened 2026-09-21
Diagnostic: find the Windows test crash (do not merge)

trypsynth:
Throwaway. Runs the desktop tests one at a time with output, so the last test name before the access violation identifies the crash. Will be closed and the branch deleted.


trypsynth:
Diagnostic done. It showed the crash was a second attempt to open an audio device after the first failed on a runner with no sound hardware. The fix is on master as 6eb07347.

PR #913, opened 2026-09-21
Update Polish readme for 1.0 and mobile apps

michaldziwisz:
## Summary

Bring `doc/readme-pl.md` up to date with the English 1.0 documentation, including the latest mobile changes in `d96114ac`.

- Translate the complete iOS/Android guide: opening documents, text/TTS modes, the reading bar, More Options, settings, keyboard shortcuts and headset controls.
- Translate all 77 entries in the 1.0 changelog. Move four previously translated entries from their old locations rather than duplicating them.
- Update the title, system requirements, features, MP3 and Windows Write formats, Ukrainian language entry, audio shortcuts and batch OCR shortcut.
- Include page numbers in the description of `=`.
- Cover the latest shared mobile table-of-contents and Find behavior, and high-contrast text on Android as well as iOS.

Only `doc/readme-pl.md` changes. No PO catalogs, application code, English documentation or other locales are modified.

## Translation details

- Preserve the distinction between recorded-audio playback speed and TTS speech rate, and between a spoken pause and visual paragraph spacing.
- Use the existing Polish interface labels. For Export, name the two actual labels: Export Document on iOS and Export As on Android.
- Explain that hiding navigation buttons removes them from screen-reader navigation, not necessarily from the display.
- Translate key names such as Space and Left/Right while retaining modifiers and command syntax. Preserve the physical Control/Command distinction on macOS.
- Keep the Polish language list in its existing Polish alphabetical order.

## Verification

- English and Polish heading levels and list structure match: 93 headings in each document.
- The two new sections preserve their list items, numbers and inline code/shortcut tokens after normalizing translated key names.
- Local negative controls detect a missing list item, wrong heading level, damaged shortcut and missing file extension. A legal translated key name remains accepted.
- Generated Polish HTML with the same Pandoc defaults and language argument used by `crates/paperback/build/docs.rs`; verified `lang=pl`, all 93 headings and all 227 inline code spans.
- `git diff --check` passes; no catalog translations changed.

No new on-device screen-reader test was performed. CI/build results will be checked on this PR; local documentation generation is not presented as an application build.


trypsynth:
Merged, thanks! One string is still untranslated in `po/pl.po`: `{} to {}`, the label of a group in the table of contents and the headings tree, shown when a book has too many entries to list under one heading (for example "Chapter 1 to Chapter 500"). Polish is human-maintained, so the machine translator will not fill it in.

PR #917, opened 2026-09-22
Update the latest simplified Chinese language

piano777wzy:
Update the latest simplified Chinese language


trypsynth:
Merged, thanks! While checking the catalog I noticed an older mistake, not from this PR. The dialog title `Choose a colour` is translated as `选择颜色选择背景颜色(&B)...`, which looks like two strings merged into one. It should probably be just `选择颜色`. Could you fix it in your next update?

PR #921, opened 2026-09-22
Build an unsigned iOS .ipa artifact in the Build Paperback workflow

LeonarddeR:
### What exists

The Build Paperback workflow (`build.yml`) produces an installable build for every platform except iOS. The Windows, Linux, macOS and Android jobs each upload their package as a workflow artifact, kept for 30 days.

The only iOS build in CI is the compile check in `ci.yml`. It builds the app without a signature to prove that the Swift and Rust sides still fit together, and then discards the result. Nothing produces an .ipa.

### What is needed

Contributors without a Mac cannot build the iOS app at all, so they cannot try their own branches on a device.

Installing does not need a Mac, only building does. Sideloading tools such as Sideloadly or AltServer run on Windows, take an unsigned .ipa, sign it with the tester's own Apple ID and install it on a connected iPhone or iPad. What is missing is the unsigned .ipa.

### What changes

A new `build-ios` job in `build.yml`, placed next to `build-android`. It:

1. Installs pandoc, so the bundled Help readmes are generated. The iOS build task only warns when pandoc is missing, and the Help screen would then be empty.
2. Builds the Rust core for device and simulator, the XCFramework and the Swift bindings with `cargo xtask ios --release`, the same step the CI check runs.
3. Builds the app with `xcodebuild` and code signing disabled, into a fixed derived data folder so the output path is known.
4. Zips the resulting `Paperback.app` inside a `Payload` folder, which is all an .ipa is.
5. Uploads it as the `paperback-ios-unsigned` artifact, kept for 30 days like the other packages.

The job is not an input to the `release` job, and the .ipa is not attached to releases. It is unsigned and only useful to someone who signs it themselves.

### How to use it

1. Open the Build Paperback run for the branch and download the `paperback-ios-unsigned` artifact.
2. Unzip it. GitHub wraps every artifact in a zip, so the .ipa is inside.
3. Install it with Sideloadly or AltServer, signed in with an Apple ID.

Two things to know:

- The build uses the same bundle ID as the App Store version, so the store app has to come off the device first, or the sideloading tool has to change the bundle ID while signing.
- With a free Apple ID the signature lasts 7 days and the tool has to refresh it. With a paid developer membership it lasts a year.


LeonarddeR:
I have tested this end to end: the artifact from this branch is now running on my iPhone, installed from Windows without a Mac. For anyone who wants to do the same, this is what worked for me.

**Getting the build**

1. Open the Build Paperback run for the branch and download the `paperback-ios-unsigned` artifact.
2. Unzip it. The `.ipa` is inside.

**Signing and installing from Windows**

I used `plumesign`, the command-line tool that ships with every [Impactor](https://github.com/claration/Impactor/releases) release (MIT licensed, no extra app on the phone). Its GUI is not usable with a screen reader, the CLI is. Download `plumesign-windows-x86_64.exe`, put it on your `PATH`, plug in the phone, and run:

```
plumesign account login -u you@example.com
plumesign sign -p paperback-ios-unsigned.ipa --apple-id --register-and-install
```

Login asks for the Apple ID password and the two-factor code. The sign command registers the phone on your account, signs the app with a free development certificate and installs it. It appends your team ID to the bundle ID by itself, so it installs next to the App Store version instead of replacing it. Press Enter at the device prompt to accept the connected phone.

A free Apple ID gives a signature that lasts 7 days. Re-running the sign command with the phone connected refreshes it.

**On the phone**

Opening the app the first time shows an "Untrusted Developer" alert. Two settings fix that, in this order:

1. Settings, General, VPN & Device Management: open the Developer App entry with your Apple ID and tap Trust.
2. Settings, Privacy & Security, Developer Mode: turn it on and let the phone restart.

**One pitfall**

If login fails with Apple error -22406 ("enter the correct password") even though the password is right, plumesign does not handle one of Apple's two password-hashing variants ([claration/Impactor#35](https://github.com/claration/Impactor/issues/35)). Changing the Apple ID password once moves the account to the variant it does handle. That is what I had to do.

The phone driver came from the Microsoft Store "Apple Devices" app. iTunes was not needed.

PR #923, opened 2026-09-22
Add a hand-written reStructuredText parser

aryanchoudharypro:
## Summary
- Adds `.rst`/`.rest` support as a new registered format (`paperback-formats`), parsed by a new `RstParser` in `paperback-core`.
- The RST-to-HTML translation is hand-written rather than wrapping the existing `rst_parser`/`rst_renderer` crates: those panic via `unimplemented!()` on constructs real documents contain (broken inline markup, Sphinx substitution references), which isn't acceptable for a file the app just needs to open. Everything here degrades instead of failing: an unrecognized directive still shows its body, and no input should ever panic it.
- Covers section titles (over/underline, first-seen-character heading levels), paragraphs, bullet/enumerated/definition/field lists (with nesting), literal blocks, a set of common directives (admonitions, image/figure, code-block, and a skip-list for toctree/include/raw/etc.), comments, transitions, block quotes, and inline markup (bold/italic/literal, roles, hyperlink references, footnote/citation refs, substitutions, bare URLs).
- Hyperlink references resolve as genuine in-book links where possible: headings and explicit anchor targets get an `id` attribute, so a cross-reference becomes a real jump-to-position link through the same `id_positions`/`resolve_link` path EPUB/HTML/CHM already use, rather than falling back to plain text. External links (`http://`, `https://`, `mailto:`) go through the same `is_external_url` classification every other parser uses.
- Output is fed through the existing shared `HtmlToText` converter, the same pipeline Markdown's parser uses, so headings/TOC/links/lists/formatting markers all come for free.

## Test plan
- [x] `cargo test -p paperback-core -p paperback-formats` — all passing (42 new RST-specific tests, including a "never panics on malformed input" battery and two real bugs found and fixed while testing against actual project changelogs: a double-backtick-vs-reference scanning bug, and footnote definitions being silently dropped as comments)
- [x] `cargo clippy -p paperback-core -p paperback-formats --all-targets` — no new warnings/errors introduced
- [x] `cargo fmt --check` — clean
- [x] Manually verified with the `pb` CLI against a hand-built sample `.rst` covering every construct above


trypsynth:
Nice work, and thanks for not wrapping the crate that panics. I tried it against a real Sphinx document (CPython's `Doc/howto/sorting.rst`), some hand-written files, and 400 bytes of random binary. Nothing panicked, unknown directives degraded as promised, and headings, lists, literal blocks and links all came through.

Two things before this goes in.

**1. A role at the start of a line is eaten as a field list.**

```rst
See the :func:`sorted` builtin, and
:func:`min`, which is smaller.
```

reads as:

```
See the sorted builtin, and
func
`min`, which is smaller.
```

The first role is fine. The second starts a line, so it matches the `:name: value` field-list shape and the role name is shown as a field name with the content left in backticks. In docutils a field list only starts a body element at the start of a block, not on a continuation line of a paragraph. This shows up all over Sphinx documents, since roles wrap onto the next line constantly.

**2. Grid tables collapse into one line.**

```rst
+------+------+
| a    | b    |
+======+======+
| 1    | 2    |
+------+------+
```

comes out as one paragraph: `+------+------+ | a | b | +======+======+ | 1 | 2 | +------+------+`. The module docs say tables are "rendered as their literal source lines", which would be readable, but the line breaks are lost, so a screen reader reads one long run of pipes and dashes. Keeping them as a literal block would at least preserve the rows.

Everything else I threw at it looked right.

PR #931, opened 2026-09-23
Finnish: updated readme and UI translation for Paperback 1.0

jkinnunen:


trypsynth:
Thanks! The catalog is complete at 965 strings and passes msgfmt, and the readme matches the English structure almost exactly. Two small things:

1. **A dropped feature bullet.** The Features list has 13 items against English's 14. The missing one is `Supports a massive array of common file formats.`, which should sit between the portable/installed line and the audio books line.
2. **A lost accelerator.** `Play &sounds on bookmarks and notes` came through as `Ilmaise kirjanmerkit ja muistiinpanot merkkiäänellä` with no `&`, so that Settings checkbox has no keyboard access. Everything else keeps its accelerator.

(The Format check failure on this PR was the Rust toolchain install failing on the runner, nothing to do with your changes. It is fixed on master now.)

jkinnunen:
Thanks for catching those! I've fixed both issues now.

trypsynth:
Merged, thanks, and both fixes look right: the Features list is complete and `Play &sounds on bookmarks and notes` has its accelerator back.

Two things have landed in the readme since you started, both today, and I'm sorry for the moving target. They're listed in https://github.com/trypsynth/paperback/discussions/910#discussioncomment-18582495: a new sub-section about a JAWS setting, and reStructuredText in the formats list plus one changelog line. There's also one new string, `Original file`. Nothing else has changed, and I'm holding readme edits from here.

